PR #5, opened 2015-02-12
Three solved problems

Hermelina:


problema2
```diff
@@ -0,0 +1,27 @@
+def identifica(array)
+  raise Exception.new("Array prea mic") if array.length < 2
+
+   largest1 = array[0]
+   largest2 = array[1]
+
+    if largest1 < largest2 then
+        temp = largest1
+        largest1 = largest2
+        largest2 = temp
```
balauru: ruby poate:

````
largest1, largest2 = largest2, largest1
````
De ce?

File: problema1 [hunk 2]
```diff
@@ -0,0 +1,15 @@
...
+#  def even_sum0(a)
+#    suma = 0
+#    a.each { |i|
+#      if i%2 == 0 then suma += i end
+#    }
+#    suma
+#  end
+
+  def even_sum(a)
+    a.find_all { |i| i%2 == 0 }.inject(0){|sum,x| sum + x }
```
balauru: (array#select)[http://ruby-doc.org//core-2.2.0/Array.html#method-i-select]

`even?` is so much nicer

problema1_test
```diff
@@ -0,0 +1,36 @@
...
+  def setup
+    # Do nothing
+  end
+
+  # Called after every test method runs. Can be used to tear down fixture information.
+  def teardown
+    # Do nothing
+  end
+
+  def test_simplu
```
balauru: The problem text in in English it will be good to preserve the language in the solutions and tests.

balauru: Even in Romanian this is not a proper [test naming](http://stackoverflow.com/questions/155436/unit-test-naming-best-practices)

problema3
```diff
@@ -0,0 +1,4 @@
+def duplicate(a)
+  a.each_with_index { |x, i|  if i != a.rindex(x) then return true end}
```
balauru: `Do not use 2 nesting loops`

The implementation of `rindex` is:
```C
static VALUE
rb_ary_rindex(int argc, VALUE *argv, VALUE ary)
{
    VALUE val;
    long i = RARRAY_LEN(ary);

    if (argc == 0) {
        RETURN_ENUMERATOR(ary, 0, 0);
        while (i--) {
            if (RTEST(rb_yield(RARRAY_PTR(ary)[i])))
                return LONG2NUM(i);
            if (i > RARRAY_LEN(ary)) {
                i = RARRAY_LEN(ary);
            }
        }
        return Qnil;
    }
    rb_scan_args(argc, argv, "1", &val);
    if (rb_block_given_p())
        rb_warn("given block not used");
    while (i--) {
        if (rb_equal(RARRAY_PTR(ary)[i], val))
            return LONG2NUM(i);
        if (i > RARRAY_LEN(ary)) {
            i = RARRAY_LEN(ary);
        }
    }
    return Qnil;
}
```

balauru:
Overall comment:

Commiting the solution and then the test is not [TDD](http://en.wikipedia.org/wiki/Test-driven_development).

problema1
```diff
@@ -0,0 +1,15 @@
+
+#  def even_sum0(a)
+#    suma = 0
+#    a.each { |i|
```
balauru: prefer `do` when the block expands multiple lines

problema1_test
```diff
@@ -0,0 +1,36 @@
+require 'test/unit'
+require_relative "problema1"
+
+class MyTest < Test::Unit::TestCase
+
+  # Called before every test method runs. Can be used to set up fixture information.
+  def setup
```
balauru: Not used code should be removed

problema3_test
```diff
@@ -0,0 +1,16 @@
+require 'test/unit'
+require_relative "problema3"
+
+class MyTest < Test::Unit::TestCase
+
```
balauru: To many empty lines.

balauru:
The project file structure is bad, please recreate it, a proper structure will be:
````
lib
 - even_sum.rb
 .....
test
 - even_sum_test.rb
 .....
Gemfile
````

See [this](http://stackoverflow.com/questions/614309/ideal-ruby-project-structure) for a longer explanation.

**Comment by balauru** [comment]
The pull request is bad, please redo it by, the preferred flow will be:

````
- fork the project
- create a branch for your solutions on your fork
- push the changes to your fork/branch
- create a pull request from your fork/branch to agilefreaks/master
```

PR #6, opened 2015-02-20
Rezolvari probleme

cipriandraghici93:
All problems are done.


rezolvari probleme.js
```diff
@@ -0,0 +1,71 @@
...
+    return !(suma2 == suma1);
+}
+
+function assert( expression, description ) { 
+	var assertStatus = expression ? 'pass' : 'fail'; 
+    console.log(description + ' : ' + assertStatus);
+}; 
+
+
+// unit testing
```
cristi-badila: This is not proper unit testing. A nice framework for unit testing in javascript is Jasmine.js. A quick google search for jasmine tutorials gave me this result: http://evanhahn.com/how-do-i-jasmine/ . Take a look at that and see if you like how it works. If you don't like it, you should try to find another framework, but as it is these are not proper tests.

rezolvari probleme.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+//problema 2
+function Get2LargestNo(vect)
+{
+    var numbers = [];
+    numbers[0] = -Infinity;
+    numbers[1] = -Infinity;
+
+    for(i=0; i < vect.length; i++)
```
balauru: What happens if the array only has 1 element, how about 2 equal elements?

rezolvari probleme.js
```diff
@@ -0,0 +1,71 @@
...
+{
+    var sum = 0;
+
+    for(i=0; i < vect.length; i++)
+    {
+        sum += vect[i];
+    }
+
+    return sum;
+}
```
cristi-badila: I think you missed a part of the problem here: ``an integer array of size N, compute the sum of all  even numbers in this array.`` Key word here is **even**

rezolvari probleme.js
```diff
@@ -0,0 +1,71 @@
...
+
+//problema 2
```
balauru: "Good code should be self explanatory"

This comments are wrong because:
- you should have renamed the function
- they are in Romanian, there is no requirement written in Romanian.

rezolvari probleme.js
```diff
@@ -0,0 +1,71 @@
...
+    {
+        if(numbers[1] < vect[i])
+        {
+            numbers[0] = numbers[1];
+            numbers[1] = vect[i];
+        }
+    }
+
+    return numbers;
+}
```
cristi-badila: Sorry to say but this implementations doesn't work. For this test data: ``[1,2,3,6,5,4]``, the result your function gives is ``[3, 6]``

rezolvari probleme.js
```diff
@@ -0,0 +1,71 @@
...
+
+//problema3
+function ThereAreDuplicate(vect)
```
balauru: The solution is incorrect, example array [2, 4, 2, 2]

rezolvari probleme.js
```diff
@@ -0,0 +1,71 @@
...
+{
+    var suma1 = SumVect(vect);
+	//console.log('suma1 : ' + suma1);
```
balauru: bad comments, probably used to debug the solution instead of making proper unit testing

rezolvari probleme.js
```diff
@@ -0,0 +1,71 @@
...
+	//console.log('vect length : ' + vect.length);
+    var suma2 = vect.length * ((vect.length + 1) / 2);
+	//console.log('suma2 : ' + suma2);
+    return !(suma2 == suma1);
+}
+
+function assert( expression, description ) { 
+	var assertStatus = expression ? 'pass' : 'fail';
```
balauru: see @cristi-badila comment on proper unit testing

balauru: Inconsistent formatting.

rezolvari probleme.js
```diff
@@ -0,0 +1,71 @@
...
+
+
+//1, 1, 3, 4, 5, 6, 7, 8, 9   false
+//1,2,3,4,5,6,7,8,9,1   false
+//1,1,1,1,1,1,1,1,1,1   false
+//1,2,3,4,5,5,7,8,9,10  false
+//1,2,3,4,5,6,7,8,9   true
+//console.log('Suma elementelor este: ' + SumVect(test1));
+//console.log('Cele mai mari 2 numere sunt: ' + Get2LargestNo(test1));
+//console.log('Exista elemente duplicate: ' + ThereAreDuplicate(test1));
```
pirvudoru: Try to avoid leaving comments in src

pirvudoru: To keep things easier to follow, you can consider separating each problem in it's on file.

rezolvari probleme.js
```diff
@@ -0,0 +1,71 @@
...
+    for(i=0; i < vect.length; i++)
+    {
+        sum += vect[i];
+    }
+
+    return sum;
+}
+
+//problema 2
+function Get2LargestNo(vect)
```
cristi-badila: Like @balauru mentioned, good code should be self explanatory, comments shouldn't be needed in the majority of cases. A part of what this means to me is that functions and variables should have very clear and descriptive names. (do not use shorthand names)

ProblemsSolved.js
```diff
@@ -0,0 +1,76 @@
...
+{
+    var pairNumbers = [];
+    pairNumbers[0] = -Infinity;
+    pairNumbers[1] = -Infinity;
+	
+	var localMaximum = -Infinity;
+	var indexOfMaximumNumber = 0;
+	
+	if(array.length < 2)
+		return 'Array length is to small. Give an array with minimum 2 pairNumbers';
```
pirvudoru: Inconsistent return type for function (string and array of numbers). 
Instead of returning string when preconditions are not met, consider "throw"-ing.

ProblemsSolved.js
```diff
@@ -0,0 +1,76 @@
...
+	
+	if(array.length == 2)
+	{
+		if(array[0] > array[1])
+		{
+			var temp = array[1];
+			array[1] = array[0];
+			array[0] = temp;
```
pirvudoru: A good approach would be not to alter the input variables when the function name does not say it does this.
Here, if you were to call:
````js
x = [ 4, 2 ];
y = GetTheLargestTwoNumer(x); // -> [ 2, 4 ];
x === y; // -> true
x; // -> [ 2, 4 ];
````

File: problem1Spec.js [hunk 13]
```diff
@@ -0,0 +1,41 @@
...
+	
+	beforeEach(function() {
+		objectProblem1 = new ProblemsSolved();
+	});
+  
+  it("Sum of even numbers in case of size of array is 0", function() {
+	testArray = [];
+	expect(objectProblem1.SumOfEvenNumberInArray(testArray)).toEqual(0);
+  });
+});
```
pirvudoru: Do you like the following approach with nesting describes ?

````js

describe('Problem1', function () { 
    var instance, subject;
	
    beforeEach(function() {
        instance = new problem1();
    });

    describe('SumOfEvenNumberInArray', function () {
        var testArray;

        beforeEach(function () {
            testArray = [];
            subject = function () {
                return instance.SumOfEventNumberInArray(testArray);
            };
        });

        describe('when array size is 0', function () {
            beforeEach(function () {
                testArray = [];
            });

           it('returns 0', function () {
              exspect(subject()).toEqual(0);
           });
        });

        describe('when array does not contain even numbers', function () {
            beforeEach(function () {
                testArray = [1, 3, 1, 1, 5];
            });

           it('returns 0', function () {
              exspect(subject()).toEqual(0);
           });
        });

        describe('when array contains even numbers', function () {
            beforeEach(function () {
                testArray = [4,4,4,4,5];
            });

           it('returns sum of event numbers', function () {
              exspect(subject()).toEqual(16);
           });
        });
    });
});

````

Do you find this easier to follow ?

File: ProblemsSolved.js [hunk 14]
```diff
@@ -0,0 +1,76 @@
+function ProblemsSolved() 
+{
+}
+
+ProblemsSolved.prototype.SumOfEvenNumberInArray = function(array)
+{
+    var sum = 0;
+    for(i=0; i < array.length; i++)
```
pirvudoru: Asigning variables without declaration will define your variable in your global namespace (for browser -> window).
So when you say here for (i = 0...) will define i on window, so you will be able to access it through  window.i

If you were to write for (var i = 0; ... ), the "i" variable will be available only in your function's scope, and not reach the global context.

ProblemsSolved.js
cipriandraghici93: No requirement specified in this problem. The only restriction was: "Do not sort the array.". I determined that the previous maximum be the first  and the current maximum to to be second, in that case. This helped me at unit testing because I knew which element is the maximum and current.

pirvudoru: I said that as a good practice.
However, above you define 

````js
pairNumbers[0] = -Infinity;
pairNumbers[1] = -Infinity;
````

Why haven't you asigned the results in that array (which you use to return result) instead of switching values in the original array ?

pirvudoru: And btw, why is there a special case for length == 2 ? Couldn't this be done in a generic way for all arrays with length > 1 ?

problem1Spec.js
cipriandraghici93: I do not know how to work professionally, but maybe it's much easier if each test should be in a separate file.

ProblemsSolved.js
cipriandraghici93: Because I used C# to implement problem first and after that I converted the code to javascript, I missed to exchange the type of variable (in that case : int i ) with "var" from javascript. But was working and  I have not observed the mistake. Ugly mistake :)

cipriandraghici93: I missed that :)

ProblemsSolved.js
cipriandraghici93: Thanks for advise. 

And yes, I can use the condition ( length > 1). Thanks you again.

PR #9, opened 2015-03-29
Problems

OvidiuBan:
version 1


Problema3.txt
```diff
@@ -0,0 +1,29 @@
...
+            else
+                if (multiply != factorial)
+                    return true;
+            return false;
+        }
+        
+          public void Arrays_WithDuplicates_ShouldFail()
+        {
+            int[] a = new int[] {2,2,6,1,3,3,5};
+            Methods method = new Methods();
```
balauru: Cred ca undeva e un `Methods` class aici nu se vede.

OvidiuBan: Ah da, am pus doar efectiv metoda care rezolva problema, Methods e o clasa
in care exista metoda ThereAreDuplicates care primeste un array. iar
Arrays_WithDuplicates_ShouldFail e din interiorul unit testing-ului

balauru: E ok,

Solutia pare valida, totusi doream sa reusesti sa foloseti git-ul si
githubul corect, dar o sa discutam despre asta, astem rezolvarile la restul
problemelor

pirvudoru:
````c#
          var list = new List<int>();

            for (var index = 1; index < 91; index++)
            {
                list.Add(index);
            }

            list.Add(92);
            list.Add(92);
            list.Add(92);
````

Care e rezultatul functiei tale pt instructiunile de mai sus (ThereAreDuplicates(list.ToArray()) ?

Comment (OvidiuBan) [inline Problema3.txt]: Pentru instructiunile de mai sus(fiind version1) returneaza ca a trecut testul pentru ca depaseste limita long-ului si produsul da 0 in ambele cazuri. La versiunea updata la care lucrez acuma, folosesc BigInteger in loc de long si se rezolva problema.

Comment (pirvudoru) [inline Problema3.txt]: :+1:  Exact ce vroiam sa aflu.

File: Problem1/MethodProblem1.cs [hunk 2]
```diff
@@ -0,0 +1,22 @@
...
+using System.Threading.Tasks;
+
+namespace SimpleProblem3
+{
+    public class MethodProblem1
+    {
+        public int SumOfEvenNumbers(int[] a)
+        {
+            int sum = 0;
+            for (int i = 0; i < a.Length; i++)
```
balauru: `foreach` ftw ca doar suntem in c# :)

Problem1/MethodProblem1.cs
```diff
@@ -0,0 +1,22 @@
...
+            {
+                if (a[i] % 2 == 0)
+                    sum += a[i];
+            }
+            return sum;
+        }
```
cristi-badila: If you like working in C# you might want to look into LINQ. Here's a solution done using it:
```
public int SumOfEvenNumbers(int[] numbers) {
  return numbers.Where(number => number % 2 == 0).Aggregate(0, (result, number) => result + number);
}
```

Problem2/UnitTestProblem2.cs
```diff
@@ -0,0 +1,25 @@
...
+            Assert.AreNotEqual(actualMax[0], int.MinValue, "Empty Array");
+            Assert.AreNotEqual(actualMax[1], int.MinValue, "Just One Element In Array");
+            int[] expectedMax = new int[] {4,2}; 
+            Assert.AreEqual(actualMax[0], expectedMax[0], "Different Max 1!");
+            Assert.AreEqual(actualMax[1], expectedMax[1], "Different Max 2!");
+           
+        }
+        
+    }
+}
```
cristi-badila: From my point of view, a good practice when writing unit tests is to keep a single assert per test. Also it looks like the first set of asserts is redundant taking into account the second one.
Here's how I would have written this test (using the ```FluentAssertions``` Nuget package):

```
[TestMethod]
public void TwoLargestNumbersAlwaysReturnsTheTwoLargestNumbersInTheGivenArray()
{
    var instance = new MethodProblem2();

    instance.TwoLargestNumbers(new[] { 0, 2, 4 }).Should().Equal(new[] { 4, 2 });
}
```

OvidiuBan: Was the first time I wrote unit tests so I was not aware if it's okay to write like this or use single assert/test. Thank you for the suggestion. I was also not certain if the array might be empty/only has one element so I added the first assets in order to treat those situations and show the cause if the test fails

Problem3/MethodProblem3.cs
```diff
@@ -0,0 +1,36 @@
...
+            }
+            gaussSum = size * (size + 1) / 2;
+            if (gaussSum != actualSum)
+
+                return true;
+            else
+                if (multiply != factorial)
+                    return true;
+            return false;
+        }
```
cristi-badila: As a coding style guideline, using multiple return statements is sometimes frowned upon. Likewise in the case of if statements with no curly brackets ```{}```. To reduce confusion, the whole ```if/else``` code block could be written as: 
```return !(gaussSum == actualSum && multiply == factorial);```

Problem2/UnitTestProblem2.cs
cristi-badila: @OvidiuBan no worries. You'll get the hang of it. As to those asserts the fact remains that you can take out the first two and no matter what items you add to the array, all the cases where the test would pass with the asserts, will pass without them and all those which would fail would fail without them, making them obsolete.

OvidiuBan: Yeah I understand that, just wanted to show the reason for the fail, mostly because I thought I'm supposed to do that, it's still a pretty territory

Problem1/MethodProblem1.cs
OvidiuBan: Reading about it now and it's really nice. Thank you for the suggestion!

Problem3/MethodProblem3.cs
OvidiuBan: Good point there, when writing that program my mind was focused more on numbers(links, differences between some arrays, what sort of conditions solve the whole case) while it being 4-5 AM on Saturday morning. Once again, thank you for the tip. I realise now I'm a big fan of this code feedback, it's a good eye-opener.

cristi-badila: @OvidiuBan I'm glad to hear that. Keep up the good work.

PR #11, opened 2015-04-06
my test

florincabuz:
Hello, here are my test files (problems + unit tests)


test.js
```diff
@@ -0,0 +1,73 @@
...
+
+    	/*
+	Given an integer array of size N, compute the sum of all even numbers in this array
+	*/
+	firstTest: function () {
+    		var array = this.numbers.slice();
+		//join the elements from array into a string
+		array = array.join();	    	
+		//searches the even numbers from string and returns the matches as an array
+		array = array.match(/\w?(\d+)*[24680]/g);
```
cristi-badila: What was you thought process when you were working on this solution? Why did you choose to get the even numbers this way instead of just iterating through the array?

jasmine.js
```diff
@@ -0,0 +1,41 @@
+describe("array numbers", function() {
+
+	it("check elements from array", function() {
+	 	expect(tests.numbers).toContain(1);
+    		expect(tests.numbers).not.toContain("1");
+    		expect(tests.numbers).not.toContain("abcd");
+  	});
```
cristi-badila: What's the point of this test?

test.js
```diff
@@ -0,0 +1,73 @@
...
+	    	var array = this.numbers.slice();
+		var tempValue = 0;
+		var result = '';
+	
+		for (var i = 0; i < array.length; i++) {
+			//store element in a temp variable
+			tempValue = array[i];
+	
+			//remove element from original array
+			var index = array.indexOf(array[i]);
```
cristi-badila: While I can't say this with 100% confidence for all Javascript implementations, internally, indexOf uses a separate loop which is forbidden by the problem requirements. e.g.: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf

test.js
florincabuz: I wanted to find something different that will solve the problem but I refactored the code

jasmine.js
florincabuz: I refactored this part of the test to. I just want to make sure that  the array has only numbers.

test.js
florincabuz: I still think about another solution without using any other temporary data structures or nested loops.

jasmine.js
```diff
@@ -0,0 +1,39 @@
+describe("array numbers", function() {
+
+	it("check elements from array", function() {
+	 	expect(tests.firstTest()).toEqual(jasmine.any(Number));
```
balauru: I expect the test  to verify that the function returns the `correct number` not `any number`

jasmine.js
```diff
@@ -0,0 +1,39 @@
...
+  	});
+
+});
+
+describe("firstTest", function() {
+
+	it("check for even numbers", function() {
+    		expect(tests.numbers).toEqual(jasmine.arrayContaining([4, 6]));
```
balauru: This is a weird assertion, you are asserting that your initial array contains 4 and 6, what is the purpose of this?

jasmine.js
```diff
@@ -0,0 +1,39 @@
...
+});
+
+describe("secondTest", function() {
+
+	it("check for string result", function() {
+    		expect(tests.secondTest()).toEqual(jasmine.any(String));
+  	});
+
+	it("string containig ' & '", function() {
+    		expect(tests.secondTest()).toEqual(jasmine.stringMatching(' & '));
```
balauru: matching the & is pointless :), I expect you to match the max and second max

test.js
```diff
@@ -0,0 +1,70 @@
+var tests = {
+
+    	numbers: [1, 5, 66, 224, 6, 22, 4, 1, 6, 8, 9, 86, 3, 4, 32, 64, 10],
+
+    	/*
+	Given an integer array of size N, compute the sum of all even numbers in this array
+	*/
+	firstTest: function () {
```
balauru: The first implementation was definitely more interesting but this one is the proper one.

test.js
```diff
@@ -0,0 +1,70 @@
...
+		var biggestNr = Math.max.apply(Math, array);
+			
+		//remove the biggest number from original array
+		var index = array.indexOf(biggestNr);
+		array.splice(index, 1);
+	
+		//get the second biggest number
+		var secondNr = Math.max.apply(Math, array);
+		
+		return biggestNr + ' & ' + secondNr;
```
balauru: You are doing 2 Math.max witch probably result in looping the array 2 times (not to mention the `indexOf`) can you solve the problem in 1 loop only?

test.js
```diff
@@ -0,0 +1,70 @@
...
+    	/*
+	Given an array of numbers having size N in which every number is between 1 and N, determine 
+	if there are any duplicates in it. You are allowed to destroy the array if you like. 
+	Do not sort the array or use bit vectors. Try to work within the array without using any other 
+	temporary data structures. Do not use 2 nesting for loops since the time complexity would be high. 
+	Try to find a solution by executing just one loop. Hint: manipulate the elements of the same array
+	as you loop through.
+	*/
+	thirdTest: function () {
+	    	var array = this.numbers.slice();
```
balauru: The number array that you give as input is not correct

The problem states: `'Given an array of numbers having size N in which every number is between 1 and N` your array contains the number 224 that means it should have at least 224 elements.

balauru:
Overall comment:

Drop the global array, each method should receive its input, all thinks will become a lot cleaner this way

florincabuz:
Hi guys!
I rebuild the second and third problem and also I used separate arrays. 
I rebuild the jasmine tests also and I hope the third problem meets the requirements.

test.js
```diff
@@ -0,0 +1,62 @@
...
+		  	}
+		}
+	    	return result;
+	},
+
+    	/*
+	Given an integer array of size N, identify the largest 2 numbers in this array. Do not sort the array.
+	*/
+    	secondTest: function () { 
+    		var array = [1, 10, 2, 9, 3, 8, 4, 7, 5, 6, 11, 4];
```
balauru: If I give you a input array of: `[-1, -3, -4]` you will return `0 & 0` a incorrect answer

jasmine.js
```diff
@@ -0,0 +1,23 @@
...
+	it("check string result", function() {
+    		expect(tests.secondTest()).toEqual(jasmine.stringMatching('11 & 10'));
+	});
+
+});
+
+describe("thirdTest", function() {
+
+	it("check string result", function() {
+    		expect(tests.secondTest()).toEqual(jasmine.stringMatching('4 5 10'));
```
balauru: the third problem requires you to determine if an array has a duplicates or not, there is no requirement to return an array

test.js
```diff
@@ -0,0 +1,62 @@
...
+	Try to find a solution by executing just one loop. Hint: manipulate the elements of the same array
+	as you loop through.
+	*/
+    	thirdTest: function () {
+    		var array = [1, 10, 2, 9, 5, 4, 4, 7, 5, 6, 3, 10, 8];
+		var result = '';
+
+		array.reduce( function(numbers, element) {
+			if (numbers.indexOf(element) < 0) {
+				numbers.push(element);
```
balauru: `Try to work within the array without using any other temporary data structures`

`numbers` is a temporary data structure, there are solution without it.

jasmine.js
```diff
@@ -0,0 +1,23 @@
+describe("firstTest", function() {
```
balauru: The test are definitely a improvement!

test.js
florincabuz: bug fixed

jasmine.js
florincabuz: So I don't need to identify them or the number of duplicates?

test.js
florincabuz: done :)

jasmine.js
florincabuz: Thanks!

test.js
```diff
@@ -0,0 +1,69 @@
...
+    		var biggestNr = -Infinity;
+    		var secondNr = -Infinity;
+
+    		for (var i = 0; i < array.length; i++) {
+    			if(array[i] > biggestNr) {
+    				secondNr = biggestNr;
+    				biggestNr = array[i];
+    			} 
+
+    			if(array[i] > secondNr && array[i] < biggestNr){
```
balauru: Ai putea sa simplifici lucrurule aici:

```
if(array[i] > biggestNr) {
    				secondNr = biggestNr;
    				biggestNr = array[i];
    			} else if(array[i] > secondNr) {
```

test.js
```diff
@@ -0,0 +1,69 @@
...
+
+    	/*
+	Given an array of numbers having size N in which every number is between 1 and N, determine 
+	if there are any duplicates in it. You are allowed to destroy the array if you like. 
+	Do not sort the array or use bit vectors. Try to work within the array without using any other 
+	temporary data structures. Do not use 2 nesting for loops since the time complexity would be high. 
+	Try to find a solution by executing just one loop. Hint: manipulate the elements of the same array
+	as you loop through.
+	*/
+    	thirdTest: function () {
```
balauru: This is finally a good solution respecting all the problem requierments

PR #12, opened 2015-05-10
Create SolutionsAndTests

dorelb:
Dorel's initial submit

### DIFF
--- a/SolutionsAndTests
+++ b/SolutionsAndTests
@@ -1,0 +1,258 @@
+function problem1(numbers)
+{
+	// safety checks, test for param null or not an array
+	if(	numbers == null)
+	{
+		return 0;
+	}
+	if(!(Object.prototype.toString.call( numbers ) === '[object Array]' ))
+	{
+		return 0;
+	}
+	// loop every element
+	var sum = 0;
+	for(i = 0; i < numbers.length; i++)
+	{
+		if(numbers[i] % 2 == 1 || numbers[i] % 2 == -1)
+		{
+			// consider even numbers only
+			sum += numbers[i];
+		}
+	}
+	return sum;
+}
+
+function problem2(numbers)
+{
+	// safety checks, test for param null or not an array
+	if(	numbers == null)
+	{
+		alert("problem1 ends-zero/null");
+		return 0;
+	}
+	if(!(Object.prototype.toString.call( numbers ) === '[object Array]' ))
+	{
+		alert("problem1 ends-zero/noarray");
+		return 0;
+	}
+	// init result array, [0]=biggest, [1]=2nd biggest
+	var result = new Array();
+	if(numbers.length >= 1)
+	{
+		result[0] = numbers[0];
+		if(numbers.length >= 2)
+		{
+			result[1] = Number.MIN_SAFE_INTEGER;
+		}
+	}
+	// loop every element
+	for(i = 1; i < numbers.length; i++)
+	{
+		if(numbers[i] > result[1])
+		{
+			// bigger than 2nd last biggest found
+			if(numbers[i] > result[0])
+			{
+				// bigger also than last biggest found, save previous biggest (as 2nd biggest), save current item (as new biggest)
+				result[1] = result[0];
+				result[0] = numbers[i];
+			}
+			else
+			{
+				// save over 2nd biggest
+				result[1] = numbers[i];
+			}
+		}
+	}
+	return result;
+}
+
+function problem3(numbers)
+{
+	// safety checks, test for param null or not an array
+	if(	numbers == null)
+	{
+		alert("problem1 ends-zero/null");
+		return 0;
+	}
+	if(!(Object.prototype.toString.call( numbers ) === '[object Array]' ))
+	{
+		alert("problem1 ends-zero/noarray");
+		return 0;
+	}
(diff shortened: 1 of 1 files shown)


SolutionsAndTests
```diff
@@ -0,0 +1,258 @@
...
+	if(	numbers == null)
+	{
+		alert("problem1 ends-zero/null");
+		return 0;
+	}
+	if(!(Object.prototype.toString.call( numbers ) === '[object Array]' ))
+	{
+		alert("problem1 ends-zero/noarray");
+		return 0;
+	}
```
cristi-badila: The previous lines are duplicated 3 times (for each problem). Wouldn't it have been easier to extract a separate function to make these checks?

SolutionsAndTests
```diff
@@ -0,0 +1,258 @@
...
+	
+	var duplicatesArray4 = [1, 2, 3, 2]
+	var duplicatesArray4Response = problem3(duplicatesArray4);
+	if(duplicatesArray4Response != true)
+	{
+		//alert("simpleArray1=>" + duplicatesArray4Response);
+		return false;
+	}
+	return true;
+}
```
cristi-badila: These are not proper unit tests. A nice framework for doing unit tests in javascript is [jasmine](http://jasmine.github.io/2.3/introduction.html).

SolutionsAndTests
```diff
@@ -0,0 +1,258 @@
...
+{
+	// safety checks, test for param null or not an array
+	if(	numbers == null)
+	{
+		return 0;
+	}
+	if(!(Object.prototype.toString.call( numbers ) === '[object Array]' ))
+	{
+		return 0;
+	}
```
cristi-badila: I don't think these safety checks are required. Given the fact that the specifications of the problem say that you are given an array, I think you should go ahead and make the calculations. As a matter of fact, these kind of pre-emptive checks might hide errors in a bigger system where for example a component should pass an array to another component but it actually passes a null value. In these cases, I think, your code **should** crash or at least raise a warning to signal that something is wrong and not just accept the invalid value.

SolutionsAndTests
```diff
@@ -0,0 +1,258 @@
...
+	{
+		//alert("withNegatives=>" + withNegativesResponse);
+		return false;
+	}
+	var repeating = [1, 1, 2, 2, 1, 2]
+	var repeatingResponse = problem1(repeating);
+	if (repeatingResponse != 3)
+	{
+		//alert("repeating=>" + repeatingResponse);
+		return false;
```
levelboy: If a test fails, how will you know which one failed?

SolutionsAndTests
```diff
@@ -0,0 +1,258 @@
...
+	}
+	if(!(Object.prototype.toString.call( numbers ) === '[object Array]' ))
+	{
+		return 0;
+	}
+	// loop every element
+	var sum = 0;
+	for(i = 0; i < numbers.length; i++)
+	{
+		if(numbers[i] % 2 == 1 || numbers[i] % 2 == -1)
```
levelboy: Problem is:'Given an integer array of size N, compute the sum of all even numbers in this array'

This if is only true for odd numbers instead of even numbers.
Even numbers are ````numbers[i]%2 == 0

SolutionsAndTests
```diff
@@ -0,0 +1,258 @@
...
+		{
+			result[1] = Number.MIN_SAFE_INTEGER;
+		}
+	}
+	// loop every element
+	for(i = 1; i < numbers.length; i++)
+	{
+		if(numbers[i] > result[1])
+		{
+			// bigger than 2nd last biggest found
```
balauru: [Good code is self explanatory](http://aspirantcoder.com/writing-self-explanatory-code/)

SolutionsAndTests
```diff
@@ -0,0 +1,258 @@
...
+	var n = numbers.length;
+	// loop every element
+	for(i = 0; i < n; i++)
+	{
+		var iteratingValue = numbers[i]; // take the value
+		// if the value was previously altered, obtain original value
+		var consideredValue = iteratingValue;
+		if(iteratingValue > n)
+		{
+			consideredValue = iteratingValue % n
```
cristi-badila: Changing what a variable represents in the middle of your code is bad practice. (before it was a value, now it's an index). Making a few renamings in your solution and removing some code makes it, to me, easier to understand and requires almost no comments:

```
function hasDuplicates(numbers) {
	var hasDuplicates = false;
	var n = numbers.length;
	for(var i = 0; i < n; i++)
	{
		var currentValue = numbers[i];
		var affectedIndex = currentValue > n ? (currentValue % n) : currentValue;
		affectedIndex--; //adjust for 0 based indexing
		if(numbers[affectedIndex] <= n)
		{
			numbers[affectedIndex] += n;
		}
		else
		{
			hasDuplicates = true;
			break;
		}
	}
	
	return hasDuplicates;
}
```

dorelb: very good observation, keeping this in mind will let the code form more readable

SolutionsAndTests
dorelb: it would worth then building up an array with failed tests, instead of a bool general test response (more alerts may not look nice if several tests fail)

PR #13, opened 2015-06-16
Solutii test

marianrabagel:
Solutii + unit testing

### DIFF
--- a/1/SumOfEvenNumbers.cs
+++ b/1/SumOfEvenNumbers.cs
@@ -8,29 +8,17 @@ namespace AgileFreaks_test1
 {
     public class SumOfEvenNumbers
     {
-        private int s;
-
-        public SumOfEvenNumbers()
-        {
-            s = 0;
-        }
-
         public int computeEvenNumbersSum(int[] array)
         {
-            s = 0;
+            int s = 0;
             for (int i = 0; i < array.Length; i++)
                 if (isEven(array[i]))
-                    addToSum(array[i]);
+                    s += array[i];
             return s;
         }
         private bool isEven(int number)
         {
             return number % 2 == 0;
-                
-        }
-        private void addToSum(int number)
-        {
-            s += number;
         }
     }
 }
--- a/1/SumOfEvenNumbers.cs
+++ b/1/SumOfEvenNumbers.cs
@@ -17,6 +17,7 @@ namespace AgileFreaks_test1
 
         public int computeEvenNumbersSum(int[] array)
         {
+            s = 0;
             for (int i = 0; i < array.Length; i++)
                 if (isEven(array[i]))
                     addToSum(array[i]);
@@ -24,9 +25,8 @@ namespace AgileFreaks_test1
         }
         private bool isEven(int number)
         {
-            if(number % 2 == 0)
-                return true;
-            return false;
+            return number % 2 == 0;
+                
         }
         private void addToSum(int number)
         {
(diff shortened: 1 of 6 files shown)


1/SumOfEvenNumbers.cs
```diff
@@ -0,0 +1,36 @@
...
+            for (int i = 0; i < array.Length; i++)
+                if (isEven(array[i]))
+                    addToSum(array[i]);
+            return s;
+        }
+        private bool isEven(int number)
+        {
+            if(number % 2 == 0)
+                return true;
+            return false;
```
pirvudoru: How do you like this:

````c#
return number % 2 == 0;
````

File: 1/SumOfEvenNumbers.cs [hunk 2]
```diff
@@ -0,0 +1,36 @@
...
+        {
+            s = 0;
+        }
+
+        public int computeEvenNumbersSum(int[] array)
+        {
+            for (int i = 0; i < array.Length; i++)
+                if (isEven(array[i]))
+                    addToSum(array[i]);
+            return s;
```
pirvudoru: You are using an instance variable that gets modified when this method is executed. If you call this method more than once, you will get invalid results.

1/SumOfEvenNumbers.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace AgileFreaks_test1
+{
+    public class SumOfEvenNumbers
+    {
+        private int s;
+
+        public SumOfEvenNumbers()
+        {
+            s = 0;
```
pirvudoru: If you indeed want that sum to be a member on an instance, make the array a hard dependency for the instance.

marianrabagel: If I would assign 0 value to my instance variable every time my method is
called it should resolve the problem of getting invalid results for
multiple calls.

2/LargestTwoNumbers.cs
```diff
@@ -0,0 +1,84 @@
...
+                checkMaxAndSecondMaxEquality();
+            }
+        }
+
+        private void checkMaxAndSecondMaxEquality()
+        {
+            if (areEqual())
+                secondMax = int.MinValue;
+        }
+        private bool areEqual()
```
pirvudoru: The name of this method does not describe the implementation.

2/LargestTwoNumbers.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+        private bool isSecondMax(int number)
+        {
+            return (number < max) && (number > secondMax);
+        }
+        private bool isMax(int number)
+        {
+            return number > max;
+        }
+        private bool isNull(int[] array)
```
OvidiuBan: Where is this method used?

3/ArrayDuplicates.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace AgileFreaks_test1
+{
+    public class ArrayDuplicates
+    {
+        public bool hasDuplicates(int[] array)
+        {
+            for (int i = 0; i < array.Length; i++)
+            {
+                if (array[array[i] % array.Length] > array.Length)
```
balauru: You are doing this calculation twice `array[i] % array.Length` you can extract in a variable

balauru:
Mersi Marian, buna treaba, check your email

PR #14, opened 2016-02-24
problems

mariusik1:


problemsSpec.js
```diff
@@ -0,0 +1,56 @@
...
+	it("for (5,1,2,3,4) should be false", function(){
+		expect(hadDuplicate(array2)).toBe(false);
+	});
+	
+	var array3=new Array(1,2,3,2,5);
+	it("for (1,2,3,2,5) should be true", function(){
+		expect(hadDuplicate(array3)).toBe(true);
+	});
+	
+	var array4=new Array(1,1,7,7,2);
```
alexandru-calinoiu: This input arrray is not valid, the size of the array is 5 (N == 5) this means that 7 is not a valid element since is not between 1 - N

problemsSpec.js
```diff
@@ -0,0 +1,56 @@
...
+	var array1=new Array(1,1,1);
+	it("for (1,1,1) it should be 0", function(){
+		expect(sumOfEvenNr(array1)).toBe(0);
+	});
+	
+	var array2=new Array(2,0,6);
+	it("for (2,0,6) it should be 8", function(){
+		expect(sumOfEvenNr(array2)).toBe(8);
+	});
+});
```
alexandru-calinoiu: How about `array3 = new Array(1)`

problemsSpec.js
```diff
@@ -0,0 +1,56 @@
+describe("First Problem", function() {
+	var array=new Array(1,2,3,4);
```
alexandru-calinoiu: You should write the prepare statement (in this case the array creation) in a the test case.

alexandru-calinoiu:
Looks good, will followup.

PR #15, opened 2016-08-07
Requested problems and test cases

Vivendus:


Problem1/even_numbers_sum.rb
```diff
@@ -0,0 +1,19 @@
+# File: even_numbers_sum.rb
+
+class EvenNumbersSum
+    def get_sum_of_even_numbers(numbers)
```
alexandru-calinoiu: Mama ce chinui ruby :)

O alternativa ar fi:

```ruby
numbers.inject { |sum, n| n.even? ? sum + n : sum }
```

sau si mai scurt

```ruby
numbers.find_all(&:even?).reduce(:+)
```

Problem2/largest_two_numbers.rb
```diff
@@ -0,0 +1,32 @@
...
+		end
+
+		firstLargeNumber = numbers[0]
+		secondLargeNumber = numbers[1]
+		
+		if firstLargeNumber < secondLargeNumber 		
+			firstLargeNumber, secondLargeNumber = secondLargeNumber, firstLargeNumber
+		end
+		
+		for index in 2..numbers.length - 1
```
alexandru-calinoiu: `numbers[2..-1].each`

Problem3/tc_array_duplicates.rb
```diff
@@ -0,0 +1,48 @@
...
+		returnedResult = array_duplicates.has_duplicates(numbers)		
+	
+		assert_equal(false, returnedResult)
+	end
+
+	
+	def test_only_duplicates
+		
+		array_duplicates = ArrayDuplicates.new
+		numbers = Array[5,5,5,5,5,5,5,5,5]
```
alexandru-calinoiu: Invalid test `Given an array of numbers having size N in which every number is between 1 and N` deci declaratia ta trebuie sa fie `numbers = [5, 5, 5, 5, 5]` (apropo nu ai nevoie de Array in fata)

Problem3/array_duplicates.rb
```diff
@@ -0,0 +1,15 @@
+# File: array_duplicates.rb
+
+class ArrayDuplicates
+	def has_duplicates(numbers)
+		numbers.each do |number|
+			index_of_element = number.abs - 1
```
alexandru-calinoiu: `number.abs` nu isi are rost vezi enuntul

Problem3/tc_array_duplicates.rb
Vivendus: Da, dar N e 9 si 5 e intre [1,9], nu vad cu ce este gresit.

Problem1/tc_even_numbers_sum.rb
```diff
@@ -0,0 +1,48 @@
...
+require "test/unit"
+
+class TestEvenNumbersSum < Test::Unit::TestCase
+	
+	def test_empty
+		
+		even_numbers_sum = EvenNumbersSum.new
+		numbers = Array[]
+		
+		sum_of_even_numbers = even_numbers_sum.get_sum_of_even_numbers(numbers)
```
cristi-badila: I'm not entirely thrilled with the naming here. It looks like there's a lot of repetition. When reading the code aloud it sounds something like: "even numbers sum get sum of even numbers with numbers". Something shorter might look like this: `even_numbers_adder.add(numbers)`.

Problem1/tc_even_numbers_sum.rb
```diff
@@ -0,0 +1,48 @@
...
+	def test_combined_even_numbers
+		
+		even_numbers_sum = EvenNumbersSum.new
+		numbers = Array[5,10,40,13,11,22,82]
+		
+		sum_of_even_numbers = even_numbers_sum.get_sum_of_even_numbers(numbers)		
+	
+		assert_equal(154, sum_of_even_numbers)
+	end
+end
```
cristi-badila: I think these specs would look much better in general if you used rspec/mini-test which are the most used frameworks in ruby for unit testing. In the future when deciding whether to use a certain framework/library I would recommend to take into account whether it is: stable, properly supported, with an active community, open/closed source.

Problem3/tc_array_duplicates.rb
alexandru-calinoiu: Ai dreptate, vroiam sa comentez la .abs

Problem1/tc_even_numbers_sum.rb
Vivendus: You are right, looks wierd, refactoring now.

Problem2/largest_two_numbers.rb
```diff
@@ -0,0 +1,32 @@
...
+			if numbers[index] > firstLargeNumber 
+			
+				secondLargeNumber = firstLargeNumber
+				firstLargeNumber = numbers[index]
+
```
cristi-badila: You seem to use an arbitrary and varying coding style in regards to white space usage. It would be nice to be consistent as it helps when reading/updating the code.

Problem1/tc_even_numbers_sum.rb
Vivendus: You are right, it looks weird. Thanks for the suggestion.

Problem2/largest_two_numbers.rb
Vivendus: Yeah, I know. Thanks for the info.

Problem2/tc_largest_two_numbers.rb
```diff
@@ -0,0 +1,60 @@
...
+	end
+	
+	def test_with_duplicates
+		
+		largest_two_numbers = LargestTwoNumbers.new
+		numbers = Array[5.9,10,12,10,78,78]
+		
+		returned_result = largest_two_numbers.get_largest_two_numbers(numbers)		
+		
+		assert_equal(Array[78,78], returned_result)
```
cristi-badila: Why did you decide to add this test case? The requirement of the problem is to find the 2 largest numbers in the array not the 2 largest entries in the array. In my opinion the result for the input in this test case should be 78 and 12.

Problem3/tc_array_duplicates.rb
```diff
@@ -0,0 +1,48 @@
+# File:  tc_array_duplicates.rb
+
+require_relative "array_duplicates.rb"
+require "test/unit"
+
+class TestArrayDuplicates < Test::Unit::TestCase
+	
+	def test_empty
+		
+		array_duplicates = ArrayDuplicates.new
```
cristi-badila: Again the naming seems of to me. When I read "array duplicates", in my mind that means maybe a data structure that contains the duplicates found in an array and in now way a class which can determine if there are any duplicates in an array. In my opinion, a better name for this class would be `DuplicateFinder`

Problem2/tc_largest_two_numbers.rb
Vivendus: > Why did you decide to add this test case?
The problem didn't state what to do in the case of duplicates. 

> The requirement of the problem is to find the 2 largest numbers in the array not the 2 largest entries in the array.
I find this hard to understand, isn't a number an entry in the array?

> In my opinion the result for the input in this test case should be 78 and 12.
You are correct, but so is the first 78 and the second one

I will add another condition to check for duplicates

cristi-badila: The text/requirement for this problem is a bit ambiguous and might be considered incomplete (which happens often in the real world). It requires that the developer thinks a bit about the way his code will be used. In my opinion if a client requested of me to write a piece of code which finds the 2 largest numbers in an array, that piece of code would return "the 2 largest numbers in the array". "78" is the same number as "78". 10, 12 and so on are different numbers. If I'm allowed to be a bit evil, [here](https://en.wikipedia.org/wiki/Number) is the definition of a number.

pirvudoru:
You can check out https://github.com/bbatsov/ruby-style-guide :+1:

Problem2/tc_largest_two_numbers.rb
Vivendus: Good point.

Problem1/even_numbers_adder.rb
```diff
@@ -0,0 +1,9 @@
+# File: even_numbers_sum.rb
+
+class EvenNumbersAdder
+  def add(numbers)
```
levelboy: You can make this a one liner

Problem2/tc_largest_two_numbers.rb
```diff
@@ -0,0 +1,51 @@
...
+    numbers = [5,7,9,10,1000,9999,55,12,-50,15000]
+      
+    returned_result = largest_two_numbers.get_numbers(numbers)		
+      
+    assert_equal(Array[15000,9999], returned_result)
+  end
+	
+  def test_with_duplicates		
+    largest_two_numbers = LargestTwoNumbers.new
+    numbers = [5,9,10,12,10,78,78]
```
levelboy: Following @cristi-badila 's comment
How would the test go with this?
[78, 78, 5,9,10,12,10]
I think it would return 78, 78... again

Problem2/largest_two_numbers.rb
```diff
@@ -0,0 +1,29 @@
+# File: largest_two_numbers.rb
+
+class LargestTwoNumbers
```
levelboy: I see you can solve this. How about having some fun with ruby and trying to make it as short as possible?

Vivendus: I see possible way by introducing lambda, which would need some documentation on my part

Problem3/duplicate_finder.rb
```diff
@@ -0,0 +1,15 @@
...
+  def has_duplicates(numbers)	
+    numbers.each do |number|					
+      if numbers[number.abs - 1] > 0 then
+        numbers[number.abs - 1] *= -1
+      else
+        return true
+      end
+    end
+		
+    return false
```
levelboy: You have an obsession :)

Problem2/tc_largest_two_numbers.rb
Vivendus: Yes, you are right. I can fix it by adding another else to check if the largest two numbers are equal
`if first_large_number == second_large_number`
and the assign
`second_large_number = number`

Problem2/largest_two_numbers.rb
levelboy: Oh, nooo... documentation :) The agony!
Just get creative
Here's my one liner :)
```
def get_largest_two_numbers(array)
  [array.delete(array.max), array.delete(array.max)]
end
```
one possibility is, now it's taken so you cannot do this:

I'm just curious how you discover ruby.

Problem3/duplicate_finder.rb
alexandru-calinoiu: with [return](https://github.com/bbatsov/ruby-style-guide#no-explicit-return)

Problem2/largest_two_numbers.rb
Vivendus: numbers.max(2)?

Problem1/even_numbers_adder.rb
```diff
@@ -2,8 +2,6 @@
 
 class EvenNumbersAdder
   def add(numbers)	
-    sum = numbers.find_all(&:even?).reduce(:+)
-		
-    return sum == nil ? 0 : sum
+    numbers.find_all(&:even?).reduce(:+).to_i
```
levelboy: Are you using the online editor in github to edit files? :)

Vivendus: My bad, I'm not near my computer right now.

levelboy: It is weird that you are doing `to_i` since all that you are operating with are integers already.
`to_i` is there just in the case there are no even numbers, so the sum is `nil`

I was thinking at something like:
`numbers.find_all(&:even?).reduce(:+) || 0`

Problem2/largest_two_numbers.rb
levelboy:
```
[7] pry(main)> array = [78, 78, 10, 5, 4].max(2)                                                                                                                                                             
=> [78, 78]
```

Problem1/even_numbers_adder.rb
levelboy: or this:
`array.find_all(&:even?).reduce(0, :+)`

Problem2/largest_two_numbers.rb
Vivendus: `numbers.length < 2 ? (raise "Array of length too small, min 2!") : numbers.uniq.max(2)`

I think this should do it

Problem1/even_numbers_adder.rb
Vivendus: I completely forgot about that operator.

alexandru-calinoiu:
Thank you.

PR #17, opened 2017-07-10
Solved problems

IliutAndrei:
@alexandru-calinoiu @levelboy


src/problem1.rb
```diff
@@ -0,0 +1,9 @@
+class Problem1
+  def sum(array)
+    unless array.nil?
```
alexandru-calinoiu: Very c# like syntax here can you write this with no branching?

_Hint: `[].sum == 0` and `conditional assignment`_

src/problem1.rb
```diff
@@ -0,0 +1,9 @@
...
+      array.sum
```
alexandru-calinoiu: The description of the problem said: `compute the sum of all __even__ numbers in this array`

src/problem2.rb
```diff
@@ -0,0 +1,9 @@
+class Problem2
+  def identify_largest_numbers(array, no_elements_identify)
```
alexandru-calinoiu: The solution asked for the first 2 numbers, no generic implementation was required.

src/problem2.rb
```diff
@@ -0,0 +1,9 @@
...
+    unless array.nil?
```
alexandru-calinoiu: Same as above see if you can drop the branching, branch prediction is a hard optimization don't make the processors live hard.

src/problem2.rb
```diff
@@ -0,0 +1,9 @@
...
+      array.max(no_elements_identify)
```
alexandru-calinoiu: Clever, but the problem required you not to sort the array, if you read the [docs](http://ruby-doc.org/core-2.4.1/Enumerable.html#method-i-max) max will sort the array.

Please implement this without sorting.

src/problem3.rb
```diff
@@ -0,0 +1,9 @@
+class Problem3
+  def contains_duplicates(array)
+    unless array.nil?
+      array.uniq.length != array.length
```
alexandru-calinoiu: If you browse the [docs](http://ruby-doc.org/core-2.4.1/Enumerable.html#method-i-max) and you will click on toggle source you will see that `uniq` implementation uses a hash, the requirement of the problem is to use no additional data structures.

test/problem1_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rspec'
+require_relative '../src/problem1'
+
+describe 'Problem1' do
+
+  array = [2,1,3,3,4]
```
alexandru-calinoiu: This is not the best way to write this variables,

Consider using [let](https://relishapp.com/rspec/rspec-core/docs/helper-methods/let-and-let)

test/problem1_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  empty_array = []
+  nil_array = nil
+
+  it 'calculate sum' do
```
alexandru-calinoiu: One of the points of `rspec` is to be able to read the specs and make sens of them.

Take a look at this alternative way of writing them, see what happens when a spec fails:

```ruby
describe SumEven do
  subject { SumEven.new.calculate(array) }

  context 'when array is empty' do
    let(:array) { [] }

    # Bonus points if you can explain how this works?
    it { is_expected.to_be zero }
  end

  context 'when array is nil'

  context 'when array is has even numbers' do
    let(:array) { [1, 2, 4, 36, 3 ] }

    it { is_expected.to eq 42 }
  end

  context 'when array has no even numbers'
end
```

_Note: This is an example unrelated to the current spec_

alexandru-calinoiu (commented):
Good start, you get points for solving the problems not for being fast.

Read the text of the problems carefully and understand that ruby's magic comes with a cost.

src/problem1.rb
IliutAndrei: Yes indeed, sorry, looks like i need new glasses, I will solve the problem asap.

src/problem2.rb
IliutAndrei: Then how do i check if the array is nil? if the function receives a nil parameter is will surely fail. Shouldn't I treat that case in the function?

IliutAndrei:
Ultimul commit nu e bun (dupa o recitire a cerintei), il voi reface.

src/problem1.rb
```diff
@@ -0,0 +1,7 @@
+class Problem1
+  def even_numbers_sum(array)
```
alexandru-calinoiu, line 2: If you know that `[].sum == 0` how will this change your solution?

src/problem3.rb
```diff
@@ -0,0 +1,12 @@
+class Problem3
+  def contains_duplicates(array)
+    return false if array.nil? or array.empty?
+
+    array_length_sum = 0;
+    (0..array.length).each do |n|
```
alexandru-calinoiu: :) the c# in you is strong, ruby has intervals, one can write`(0..array.lenght).sum`

test/problem3_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  subject { Problem3.new.contains_duplicates(array) }
+
+  context 'when array contains duplicates' do
+    let(:array) { [2,1,3,3,4] }
+
+    it { is_expected.to be true }
+  end
+
+  context 'when array contains only duplicates' do
+    let(:array) { [1,1,1,1] }
```
alexandru-calinoiu: Try to change this to `let(:array) { [1, 1, 1] }`

test/problem1_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'rspec'
+require_relative '../src/problem1'
+
+describe Problem1 do
```
alexandru-calinoiu, line 4: Specs look much better :+1:

alexandru-calinoiu (changes requested):
Better, problem 3 is not yet solved, try one more time.

src/problem1.rb
IliutAndrei: In case the array is nil, i have to return 0, otherwise the 'select' function does not take nil as a parameter => error. And to return nil from a function is bad.

src/problem3.rb
IliutAndrei: I will refactor that.

test/problem3_spec.rb
IliutAndrei: Done

IliutAndrei:
Acuma testele sunt ok si cu noul array de test [1,1,1] la problema 3.

src/problem2.rb
```diff
@@ -0,0 +1,18 @@
+class Problem2
+  def identify_largest_numbers(array)
+    return [] if array.nil? or array.empty?
```
pirvudoru: Is an empty array a valid output for this function? By name it has to return largest (2) numbers. In this case it cannot accomplish it's goal because some preconditions are not met.

test/problem1_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  context 'when array has only 1 number' do
+    let(:array) { [2] }
+
+    it { is_expected.to eq 2 }
+  end
+
+  context 'when array is nil' do
+    let(:array) { nil }
+
+    it { is_expected.to eq 0}
```
pirvudoru: Does not seem fair to me. The problem states that you have an array of size N, so it won't be null. 
Also sum of nil to equal 0 seems odd.

test/problem3_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  context 'when array is empty' do
+    let(:array) { [] }
+
+    it { is_expected.to be false }
+  end
+
+  context 'when array is nil' do
+    let(:array) { nil }
+
+    it { is_expected.to be false }
```
pirvudoru: Why guard against nils ? Problems state that you are given an array of size N

test/problem2_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  context 'when array is empty' do
+    let(:array) { [] }
+
+    it { is_expected.to eq [] }
+  end
+
+  context 'when array is nil' do
+    let(:array) { nil }
+
+    it { is_expected.to eq [] }
```
pirvudoru: Why guard against nils ? Problems state that you are given an array of size N

src/problem1.rb
```diff
@@ -0,0 +1,7 @@
...
+    return 0 if array.nil?
+    sum = array.select(&:even?).reduce(:+)
+    sum == nil ? 0 : sum
```
pirvudoru: This can be furthure simplified. Check @alexandru-calinoiu's comment. Also no need to guard for nil. Problems state that you are given an array of size N

test/problem3_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'rspec'
+require_relative '../src/problem3'
+
+describe Problem3 do
```
pirvudoru, line 4: Add a case for `[1,1,4,4]`

pirvudoru (changes requested):

src/problem2.rb
IliutAndrei: Indeed, refactoring now.

src/problem1.rb
IliutAndrei: Done

test/problem3_spec.rb
IliutAndrei: Yes, the test fails, I will try to come up with a better implementation for this problem.

test/problem2_spec.rb
IliutAndrei: Indeed, I will remove the nil check

src/problem2.rb
```diff
@@ -0,0 +1,18 @@
+class Problem2
+  def identify_largest_numbers(array)
+    raise "Array of length too small, min 2!" if array.length < 2
+
+    max1, max2 = -1
```
alexandru-calinoiu, line 5: This is assuming that there are no negative numbers in the array, `integers` can have negative values.

alexandru-calinoiu (approved):
Problem 2 still had an issue

alexandru-calinoiu:
Will close this and followup on email.

PR #18, opened 2017-07-13
Proposed solution for problems

cristicatalan:


cristicatalan:
Please review code. Thanks

problems/lib/problem1.rb
```diff
@@ -0,0 +1,7 @@
+class Problem1
+  def sum_even_numbers(array)
+    sum = 0
```
alexandru-calinoiu: Oh but ruby is so much beautiful can you do this on one line? 

_Hint, checkout `select`, and `[].sum == 0)_

problems/lib/problem1.rb
```diff
@@ -0,0 +1,7 @@
+class Problem1
+  def sum_even_numbers(array)
+    sum = 0
+    array.each {|num| sum+=num if num.even?}
```
alexandru-calinoiu: How will you guard against `nil` arrays?

problems/lib/problem2.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      if (num>  max1)
+        max2 = max1
+        max1 = num
+      else
+        max2 = num
+      end
+    end
+
+    maximums = [max1, max2]
```
alexandru-calinoiu: Unused variable assigment.

alexandru-calinoiu (commented):
Looks good, there are some minor estetic changes but overall is :+1:

problems/lib/problem1.rb
cristicatalan: I refactored code to use select and sum.

problems/lib/problem1.rb
cristicatalan: You can guard against nil by raising an exception early, at the top of the method like this:
`raise "Array must not be nil." if array.nil?`
NB: I did's check for nil because the problem (all three in fact) state "Given an integer array of size N". A nil array cannot have a size, so I just assumed it is not.

problems/lib/problem2.rb
cristicatalan: Indeed. I just overlooked it and erroneously commited code. I refactored.

problems/lib/problem2.rb
```diff
@@ -0,0 +1,26 @@
...
+  def get_largest_two_numbers(array)
+    raise "Array must have at least two elements." if array.size < 2
+
+    if (array[1] > array[0])
+      max1 = array[1]
+      max2 = array[0]
+    else
+      max1 = array[0]
+      max2 = array[1]
+    end
```
pirvudoru: Can't we initialize with something else to avoid this big chunk of code and incorporate it in our logic ?

problems/lib/problem2.rb
```diff
@@ -0,0 +1,26 @@
+class Problem2
+  def get_largest_two_numbers(array)
+    raise "Array must have at least two elements." if array.size < 2
```
pirvudoru: Prefer raising typed errors. `ArgumentError`

problems/lib/problem3.rb
```diff
@@ -0,0 +1,14 @@
+class Problem3
+  def has_duplicate(array)
+    len = array.size
+
+    for i in (0..len-1)
+      relocated_index = array[i] % len
+      return true if array[relocated_index] > len
```
pirvudoru: Prefer having 1 return per method -> at the end of the method. Also `return` keyword is rarely -> never used https://github.com/bbatsov/ruby-style-guide

pirvudoru (commented):
Cosmetic changes, but looking good

problems/lib/problem1.rb
```diff
@@ -0,0 +1,5 @@
+class Problem1
```
levelboy: Why do we need a class? 
What approach would you take without creating a new class for this?

problems/lib/problem2.rb
```diff
@@ -0,0 +1,26 @@
+class Problem2
+  def get_largest_two_numbers(array)
```
levelboy, line 2: Yeah... it works...
But it's one big chunk of code that if converted in C wouldn't look much different :(
Ruby should be pleasant to the eye, actually all code should be...

As Kent Beck says : "Make it work, make it right, make it fast"
How would you refactor this code to be more readable, with less duplication in concepts and with less assignments

problems/test/problem1_test.rb
```diff
@@ -0,0 +1,26 @@
+require "minitest/autorun"
+require_relative "../lib/problem1"
+
+class Problem1Test < MiniTest::Unit::TestCase
+  def setup
+    @problem1 = Problem1.new
```
levelboy: Tests should give you an insight on how the code will be used. 
Would you instantiate a class, to use a method in it just to get the sum of even numbers in a given array?

problems/test/problem1_test.rb
```diff
@@ -0,0 +1,26 @@
+require "minitest/autorun"
+require_relative "../lib/problem1"
+
+class Problem1Test < MiniTest::Unit::TestCase
+  def setup
+    @problem1 = Problem1.new
+  end
+
+  def test_given_empty_array_sum_is_zero
```
levelboy: I like the name of all the tests.
Have you tried using Minitest's expectations syntax? 
They'll become even nicer as it's closer to normal speech

problems/lib/problem2.rb
cristicatalan: I agree and I changed code.

problems/lib/problem2.rb
cristicatalan: The block of code can be refactored like this:
`max1, max2 = array[0] > array[1] ? array[0..1] : array[0..1].reverse`
Although this approach seems not to be recommended, after I scanned the [Ruby coding style](https://github.com/bbatsov/ruby-style-guide) you provided on your next comment.

> Avoid the use of parallel assignment for defining variables. Parallel assignment is allowed when it is the return of a method call, used with the splat operator, or when used to swap variable assignment. Parallel assignment is less readable than separate assignment.

Next, the code can be extracted into a method and then use parallel assignment. I chose to commit code as in [here](https://github.com/cristicatalan/Test/commit/8b28b7afea012539b24cd08c06c4b32870e93359), just because of the previous explanation.
```
max1, max2 = get_largest_first_two_numbers(array)
def get_largest_first_two_numbers(array)
  array[0] > array[1] ? array[0..1] : array[0..1].reverse
end
```

Nevertheless I think extract into a method is not necessary in this case and code is readable enough without it. What do you think?

problems/lib/problem3.rb
cristicatalan: Refactored code in a way to keep it as concise as before and avoid two returns.

problems/lib/problem1.rb
cristicatalan: Two alternatives would do the job:
1. Monkey patching of the Array class, though I don't like it and definitely wouldn't use it in this case, since I lose the default sum method:
```
def Array
  def sum(array)
    array.select(&:even?).sum
  end
end
```

2. Add a module with a module level method. I amended code to use this approach.

problems/lib/problem2.rb
cristicatalan: The first part when initializing max1 and max2 variables can be refactored as per my previous [reply](https://github.com/Agilefreaks/Test/pull/18/files/6994d074a8a8d329459c390cee38354ed514af65#r127580557) to @pirvudoru.

The block of code in the loop can be refactored as such:
`max1, max2 = num > max1 ? [num, max1] : [max1, num]`
Although not sure this approach would be acceptable, since it basically would be possible to do one additional unrequired assignment of max1 in max1, max2 = [max1, num], compared to the current code. On top of that seems that [Ruby coding style](https://github.com/bbatsov/ruby-style-guide) advises against this kind of usage for parallel assignment.
Additionally it can be further extracted in a method.

Commit is [here](https://github.com/cristicatalan/Test/commit/f310106ead40015e33a3b44fc1463ac7a4d2ddb6). To me it is readable enough without extract into a method. What do you think?

problems/test/problem1_test.rb
cristicatalan: I used the approach to create and use a module level method instead and I committed code.

problems/test/problem1_test.rb
cristicatalan: Expectations definitely look nicer. I changed tests to use them.

alexandru-calinoiu:
All good, followed up via email.

PR #19, opened 2017-07-19
Initial commit

hyteoo:


agile_problem1_test.exs
```diff
@@ -0,0 +1,21 @@
+ExUnit.start
+
+defmodule Problem1Test do
+	use ExUnit.Case
+	
+	test "Problem 1: positive test 1" do
```
alexandru-calinoiu, line 6: Unit tests should have a more explicit naming, like a good writeup about this is [here](http://osherove.com/blog/2005/4/3/naming-standards-for-unit-tests.html), boils down to [UnitOfWork__StateUnderTest__ExpectedBehavior]

agile_problem1_test.exs
```diff
@@ -0,0 +1,21 @@
...
+		assert Problem1.sum_even([1,2,3,4,5,6,7,8,9]) == 20
+	end
+	test "Problem 1: positive test 2" do
+		assert Problem1.sum_even([-30, 0, 1000]) == 970
+	end
+	test "Problem 1: negative test" do
+		assert Problem1.sum_even([-5, 13, -7, 99]) == 0
+	end
+	test "Problem 1: one element test" do
```
alexandru-calinoiu, line 15: This last 2 test are better named.

agile_problem1.exs
```diff
@@ -0,0 +1,28 @@
+defmodule Problem1 do
+	import Integer
+	
+	#Main public function that sums up the even numbers
+	def sum_even(), do: 0
```
alexandru-calinoiu, line 5: Does not make sens to call even_sum with arity 0, the problem states that we want to sum of a list so there should be an arity of 1 at minimum.

agile_problem1.exs
```diff
@@ -0,0 +1,28 @@
...
+	def sum_even(list), do: do_sum_even(list, 0)
+	
+	
+	#Private functions ---------------------------------------
+	
+	defp do_sum_even([], acc), do: acc
+	
+	defp do_sum_even([head | tail], acc) do
+	    if is_even(head) do
```
alexandru-calinoiu, line 14: This is indeed a way to solve this problem, very c like, but it would have been nice to see the elixir way to do it.

This entire module can be replaced by a simple function that uses pipes and the Enum module.

agile_problem2.exs
```diff
@@ -0,0 +1,32 @@
+defmodule Problem2 do
+	def largest_2_numbers(), do: -2
+	
+	def largest_2_numbers(list) do
+		#First check that we have at least 2 elements. Otherwise it's useless
+		if Enum.count(list)<2 do
+			-1
```
alexandru-calinoiu, line 7: In elixir your will use [atoms](https://elixir-lang.org/getting-started/basic-types.html#atoms) to represent this, they have the benefit of being more readable and of course constants.

agile_problem2.exs
```diff
@@ -0,0 +1,32 @@
+defmodule Problem2 do
+	def largest_2_numbers(), do: -2
+	
+	def largest_2_numbers(list) do
+		#First check that we have at least 2 elements. Otherwise it's useless
+		if Enum.count(list)<2 do
+			-1
+		else
+			#make the list have unique values.
+			ulist = Enum.uniq(list)
```
alexandru-calinoiu, line 10: This is a inefficient way to solve this problem, `uniq` is probably O(log n) and `max` is O(n). You can solve this with O(n).

Not sure why you used unique, the 2 largest numbers in [1, 2, 2] are 2 and 2, it happens that they are the same but that is because of the input, the problem did not stated to find the unique max.

agile_problem2.exs
```diff
@@ -0,0 +1,32 @@
...
+			
+			#Find the max and delete it from the list
+			max1 = Enum.max(ulist)
+			ulist2 = List.delete(ulist,max1)
+			
+			#Find the next max
+			max2 = Enum.max(ulist2)
+			
+			{max1, max2}
```
alexandru-calinoiu, line 19: :+1: Proper way to return a result like this in a `tuplet`

agile_problem2.exs
```diff
@@ -0,0 +1,32 @@
+defmodule Problem2 do
+	def largest_2_numbers(), do: -2
```
alexandru-calinoiu, line 2: Same comment as for problem 1, + same comment for the usage of a arbitrary number instead of an atom.

agile_problem3.exs
```diff
@@ -0,0 +1,25 @@
+defmodule Problem3 do
+	def check_duplicates(), do: -2
+	
+	def check_duplicates(list) do
+		#First check that we have at least 2 elements. Otherwise it's useless
+		if Enum.count(list)<2 do
+			-1
+		else
+			if list == Enum.uniq(list) do
```
alexandru-calinoiu, line 9: Nice try but this is cheating, `Enum.uniq` uses [maps in it's implementation](https://github.com/elixir-lang/elixir/blob/v1.5.0/lib/elixir/lib/enum.ex#L2594), you where not suppose to use any extra data structures.

The same comment of using numbers vs atoms apply.

agile_problem3_test.exs
```diff
@@ -0,0 +1,22 @@
...
+	use ExUnit.Case
+	
+	test "Problem 3: positive test" do
+		assert Problem3.check_duplicates([21, 13, -20, 4, 20, 93, 100, -19, -95, -83, 3, -3, 42, -26, 100, 42, -52, 46, -91, -23]) == 1
+	end
+	test "Problem 3: two element test" do
+		assert Problem3.check_duplicates([1,2,3,4,5,6]) == 0
+	end
+	test "Problem 3: negative test - one element" do
+		Problem3.check_duplicates([-30, -15, -122, 0, 1001]) == 0
```
alexandru-calinoiu, line 13: Not a valid use case, the problem states `Given an array of numbers having size N in which every number is between 1 and N`

You don't have negative numbers
If you have the number 1001 you need to have an array with at least 1001 elements.

alexandru-calinoiu (changes requested):
Problem 1, 0.9 points (it can be solved in a more elixir fashion)
Problem 2, 1 point, it is solved inefficiently, it should be solved in O(n)
Problem 3, 0 points it does not respect the requirements.

alexandru-calinoiu:
Closing this one as we settled it offline :)

PR #20, opened 2017-09-16
rezolvare probleme test

MihailDamian:
am creat branch-uri pentru fiecare problema in parte, dintr-o greseala minora nu le-am creat din main si aproape ca nici nu se observa ca e branch separat, si la problema 2 am urcat de 2 ori cu comment "red", a 2-a oara ar fi green . In rest sper sa fie bine, astept un raspuns, si observatii :), mersi.


problem1/problem1.rb
```diff
@@ -0,0 +1,16 @@
+class Problem1
+	attr_accessor :result,:array
+
+	def  initialize(*args)
+		@array=args;
+	end
+
+	def SumEvenNumbers
+		@result=0;
+		for i in 0..(@array.size-1)
```
alexandru-calinoiu, line 10: This is very c# like, search the [array docs](http://ruby-doc.org/core-2.4.1/Array.html) can you find a more ruby like method to solve this?

problem1/problem1.rb
```diff
@@ -0,0 +1,16 @@
+class Problem1
+	attr_accessor :result,:array
+
+	def  initialize(*args)
+		@array=args;
+	end
+
+	def SumEvenNumbers
```
alexandru-calinoiu, line 8: Upercase is very atypical for method names in ruby.

problem2/problem2.rb
```diff
@@ -0,0 +1,23 @@
...
+	attr_accessor :result,:array
+
+	def  initialize(*args)
+		@array = args;
+	end
+
+	def Largest2Numbers
+		@result=[0,0];
+		for i in 0..(@array.size-1)
+			if(!@result.include? @array[i])
```
alexandru-calinoiu, line 11: This is not a great check, `include?` is O(n) so your implementation now is O(n^2), do you really need it?

problem2/problem2_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "./problem2"
+describe Problem2 do
+	it "Largest 2 Numbers" do
```
alexandru-calinoiu, line 3: How about test cases with an array of 1 element? or an array with that has the first and second max duplicated (Ex: [4, 4, 1, 3]) ?

problem3/problem3_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "./problem3"
+describe Problem3 do
+	it "Dublicate Numbers" do
+		example3 = Problem3.new(1,3,1,4,3,4,8,5)
+		example3.DublicateNumbers;
+		expect(example3.array).to  eql([1,3,4]);
```
alexandru-calinoiu, line 6: Unfortunately this is not correct, the expected result is a boolean, read the requirement again and update your solution.

problem2/problem2.rb
```diff
@@ -0,0 +1,23 @@
+class Problem2
+	attr_accessor :result,:array
+
+	def  initialize(*args)
+		@array = args;
+	end
+
+	def Largest2Numbers
```
alexandru-calinoiu, line 8: How about returning the result via the method? storying it in a instance variable makes you implementation more complicated, state mutation is usually not a good idea.

Your function should try to be [pure](https://en.wikipedia.org/wiki/Pure_function)

problem1/problem1.rb
```diff
@@ -0,0 +1,16 @@
+class Problem1
+	attr_accessor :result,:array
```
alexandru-calinoiu, line 2: Do `array` and `result` need to writable from outside the class? If not consider updating your class signature accordingly.

problem1/problem1_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "./problem1"
+describe Problem1 do
+	it "Sum Even Numbers" do
```
alexandru-calinoiu, line 3: Take a quick look at [better specs](http://www.betterspecs.org/) and take a stab at improving your specs.

alexandru-calinoiu (commented):
So far we have a incomplete c# like implementation, some more work is required.

Please review my comments.

problem1/problem1.rb
```diff
@@ -0,0 +1,16 @@
...
+
+	def  initialize(*args)
+		@array=args;
+	end
+
+	def SumEvenNumbers
+		@result=0;
+		for i in 0..(@array.size-1)
+			if @array[i] % 2 == 0 
+				@result += @array[i]
```
pirvudoru, line 12: Can you write it without using `if` ?

problem1/problem1.rb
pirvudoru: Ruby style guide: https://github.com/bbatsov/ruby-style-guide

problem2/problem2_spec.rb
```diff
@@ -0,0 +1,8 @@
...
+		example1 = Problem2.new(11,2,3,4,12,3,14)
+		example1.Largest2Numbers;
```
pirvudoru, line 5: Usually method names denote an action. Ex: `find_largest_two_numbers`.

problem2/problem2_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "./problem2"
+describe Problem2 do
+	it "Largest 2 Numbers" do
+		example1 = Problem2.new(11,2,3,4,12,3,14)
```
pirvudoru, line 4: All the problem state `Given an array`. This signature does not accept an array. In real life we would not do variable arguments length to pass in arrays since we don't actually know the number of elements in the array. Using the `splat operator` will only introduce computational / maintainability overhead in this case.

problem2/problem2.rb
```diff
@@ -0,0 +1,23 @@
...
+		for i in 0..(@array.size-1)
+			if(!@result.include? @array[i])
+				if @array[i] > @result[0] 
+					if @result[1] < @result[0]
+						@result[1] = @result[0]
+					end
+					@result[0] = @array[i]
+				elsif @array[i] > @result[1]
+					@result[1] = @array[i]
+				end
```
pirvudoru, line 19: Can we simplify these nested ifs ?

problem3/problem3.rb
```diff
@@ -0,0 +1,31 @@
...
+			if(@array[i]==1)
+				array.push(i+1);
+			end
+		end
+		for i in 0.. @array.size-1
+			if @array[i] == '-chuck-norris-'
+				@array.shift(i+1);
+				break;
+			end
+		end
```
pirvudoru, line 29: Interesting solution, but we need a boolean as a result. The `Array.new` is new data structure. You could have filled the array another way to avoid instantiating a new array.

pirvudoru (changes requested):

alexandru-calinoiu:
Unfortunately I have to consider this a failed try.

PR #21, opened 2017-09-18
Constantin Capatina test solution

drqCode:
These are my solutions to your test problems. I hope it matches your expectations.

Have a good day! :)

### DIFF
--- a/src/problem1.rb
+++ b/src/problem1.rb
@@ -1,0 +1,6 @@
+
+class Problem1
+  def self.sum_of_even_numbers(numbers)
+    numbers
+  end
+end
--- a/src/problem1.rb
+++ b/src/problem1.rb
@@ -4,9 +4,7 @@
 class Problem1
   # @param [Array] numbers
   def self.sum_of_even_numbers(numbers)
-    return 0 if numbers.nil? || numbers.empty?
-    sum = 0
-    numbers.each { |x| sum += x.even? ? x : 0 }
-    sum
+    return 0 if numbers.nil?
+    numbers.select {|x| x.even?}.sum
   end
 end
--- a/src/problem1.rb
+++ b/src/problem1.rb
@@ -1,6 +1,12 @@
 
+# Given an integer array of size N,
+# compute the sum of all even numbers in this array
 class Problem1
+  # @param [Array] numbers
   def self.sum_of_even_numbers(numbers)
-    numbers
+    return 0 if numbers.nil? || numbers.empty?
+    sum = 0
+    numbers.each { |x| sum += x.even? ? x : 0 }
+    sum
   end
 end
--- a/src/problem2.rb
+++ b/src/problem2.rb
@@ -4,7 +4,7 @@
 class Problem2
   # @param [Array] numbers
   # @return [Array]
-  def self.get_largest_two_of(numbers)
+  def self.largest_two_of(numbers)
     return nil if numbers.nil? || numbers.empty?
     return numbers if numbers.size == 1
     first = numbers[0]
--- a/src/problem2.rb
+++ b/src/problem2.rb
@@ -1,6 +1,22 @@
 
+# Given an integer array of size N,
+# identify the largest 2 numbers in this array. Do not sort the array.
 class Problem2
-  def self.get_largest_two_numbers(numbers)
-    numbers
+  # @param [Array] numbers
+  # @return [Array]
+  def self.get_largest_two_of(numbers)
+    return nil if numbers.nil? || numbers.empty?
+    return numbers if numbers.size == 1
+    first = numbers[0]
+    second = numbers[1]
+    (2..numbers.size - 1).each do |i|
+      if numbers[i] > first
+        second = first
+        first = numbers[i]
+      elsif numbers[i] > second
+        second = numbers[i]
+      end
+    end
+    [second, first]
   end
 end
--- a/src/problem2.rb
+++ b/src/problem2.rb
@@ -1,0 +1,6 @@
+
+class Problem2
+  def self.get_largest_two_numbers(numbers)
+    numbers
+  end
+end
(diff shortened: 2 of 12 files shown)


src/problem1.rb
```diff
@@ -0,0 +1,12 @@
+
+# Given an integer array of size N,
+# compute the sum of all even numbers in this array
+class Problem1
+  # @param [Array] numbers
+  def self.sum_of_even_numbers(numbers)
+    return 0 if numbers.nil? || numbers.empty?
```
pirvudoru: Assume you won't be getting any `nil` input as per problem description. Do you need the `empty?` check?

src/problem1.rb
```diff
@@ -0,0 +1,12 @@
...
+    sum = 0
+    numbers.each { |x| sum += x.even? ? x : 0 }
+    sum
```
pirvudoru: Good. But can be better. Can you write it as 1 liner ?

test/problem1_spec.rb
```diff
@@ -0,0 +1,37 @@
+
+require 'rspec'
+require '../src/problem1'
+
+describe Problem1 do
+  describe '.sum_of_even_numbers' do
+    context 'when is null array' do
```
pirvudoru, line 7: When refining contexts, you can do additional setup using `let`.

test/problem1_spec.rb
```diff
@@ -0,0 +1,37 @@
+
+require 'rspec'
+require '../src/problem1'
+
+describe Problem1 do
+  describe '.sum_of_even_numbers' do
```
pirvudoru, line 6: When describing method calls, you can use `subject` to avoid repeating whole method call.

test/problem2_spec.rb
```diff
@@ -0,0 +1,42 @@
+
+require 'rspec'
+require '../src/problem2'
+
+describe Problem2 do
+  describe '.get_largest_two_of' do
+    context 'when is null array' do
+      it 'returns nil' do
+        expect(Problem2.get_largest_two_of(nil)).to eql(nil)
```
pirvudoru: You can leave out `get` from method names https://github.com/bbatsov/ruby-style-guide#accessor_mutator_method_names

test/problem3_spec.rb
```diff
@@ -0,0 +1,42 @@
+
+require 'rspec'
+require '../src/problem3'
+
+describe Problem3 do
+  describe '.duplicates?' do
+    context 'when is null array' do
+      it 'returns false' do
+        expect(Problem3.duplicates?(nil)).to be false
```
pirvudoru, line 9: :+1: method name

pirvudoru (commented):
Few minor comments, but overall :+1:

src/problem1.rb
drqCode: Yep

src/problem1.rb
drqCode: I assumed that the function will be used by unknown provider... so that's why I wrote those checks.

test/problem1_spec.rb
drqCode: I read something, but I have been running out of time :(

src/problem1.rb
drqCode: One-liner is pretty new to me, but for now I don't see it very clean...

alexandru-calinoiu: Check the docs for [array](http://ruby-doc.org/core-2.4.1/Array.html) maybe you can see something usefull there.

src/problem2.rb
```diff
@@ -0,0 +1,22 @@
...
+# identify the largest 2 numbers in this array. Do not sort the array.
+class Problem2
+  # @param [Array] numbers
+  # @return [Array]
+  def self.largest_two_of(numbers)
+    return nil if numbers.nil? || numbers.empty?
+    return numbers if numbers.size == 1
+    first = numbers[0]
+    second = numbers[1]
+    (2..numbers.size - 1).each do |i|
```
alexandru-calinoiu, line 12: Very c way to iterate an array in ruby, can you think of a better cleaner way?

TIP: you should not be required to access elements via their index, just have the element available.

src/problem2.rb
```diff
@@ -0,0 +1,22 @@
+
+# Given an integer array of size N,
+# identify the largest 2 numbers in this array. Do not sort the array.
+class Problem2
+  # @param [Array] numbers
+  # @return [Array]
+  def self.largest_two_of(numbers)
+    return nil if numbers.nil? || numbers.empty?
```
alexandru-calinoiu, line 8: This checks are not actually required, see [yagni](https://www.martinfowler.com/bliki/Yagni.html)

test/problem2_spec.rb
```diff
@@ -0,0 +1,42 @@
+
+require 'rspec'
+require '../src/problem2'
+
+describe Problem2 do
+  describe '.largest_two_of' do
+    context 'when is null array' do
```
alexandru-calinoiu, line 7: `null` is not something that ruby has, did you mean nil?

src/problem3.rb
```diff
@@ -0,0 +1,23 @@
...
+# Do not sort the array or use bit vectors.
+# Try to work within the array without using any
+# other temporary data structures.
+# Do not use 2 nesting for loops since the time
+# complexity would be high.
+# Try to find a solution by executing just one loop.
+class Problem3
+  # @param [Array] numbers
+  # @return [boolean]
+  def self.duplicates?(numbers)
```
alexandru-calinoiu, line 15: Really liked the fact that you used `?` in the method name, that is a cool ruby naming convention.

src/problem1.rb
drqCode: Ok, so I managed to do something shorter and cleaner

alexandru-calinoiu (commented):
Over all a very good solution, please take a look at my comments.

test/problem2_spec.rb
drqCode: I've used null because I think that Unit Testing should not consider how the language is built. So i think 'null' is more understable, but as Martin Fowler says, 'yagni'...

PR #22, opened 2017-11-18
Feature branch - Problem1

intellinor:


src/problem1.rb
```diff
@@ -0,0 +1,17 @@
+class Problem1
+  def sumOfAllEvenNumbers(values)
+    response = 0
+    if(!values.nil? && (values.instance_of?(Array)))
```
alexandru-calinoiu: There is no need for this checks, the problem text is very clear that you can only receive a an array of integers

alexandru-calinoiu: This is usually consider a bad practice, read more about it here https://www.martinfowler.com/bliki/Yagni.html

src/problem1.rb
```diff
@@ -0,0 +1,17 @@
...
+      if(values.all? {|i| i.is_a?(Integer)})
+        values.select {|i| i % 2 == 0}.each {|number|
```
alexandru-calinoiu: Is there a [better](http://ruby-doc.org/core-2.2.0/Integer.html#method-i-even-3F) way to express that a number is even in ruby?

tests/problem1Tests.rb
```diff
@@ -0,0 +1,50 @@
...
+    @testDataInteger2 = (1..100).to_a
+    @testDataNotInteger1 = %w(1 2 3 4)
+    @testDataNotInteger2 = [1,nil,3,4,5,6]
+  end
+
+  def teardown
+
+  end
+
+  def test_hasExpectedSumResult1
```
alexandru-calinoiu: In a good test name at a minimum you should have:

- specify the state (context) of the test
- specify the expected behavior

src/problem1.rb
```diff
@@ -0,0 +1,17 @@
...
+          response = response + number;
```
alexandru-calinoiu: Check the array documentation and figure out if there is a better way to sum up elements in an array

alexandru-calinoiu (changes requested):
Ruby is a very versatile language, it was force here to take the shape of same bad c# code :).

Try to express the solution more naturally and minimal.

src/problem1.rb
```diff
@@ -0,0 +1,5 @@
+class Problem1
+  def sumOfAllEvenNumbers(values)
+    return values.select {|i| i.even?}.reduce(0, :+)
```
alexandru-calinoiu, line 3: Much better 🥇 

Just out of curiosity why use `reduce` instead of `sum`?

tests/problem1Tests.rb
```diff
@@ -0,0 +1,37 @@
...
+require '../src/problem1'
+
+class Problem1Tests < Test::Unit::TestCase
+
+  def setup
+    @problem1 = Problem1.new
+    @randomIntegerTestData = [1,4,7,4,10,6]
+    @orderedIntegerTestData = (1..100).to_a
+    @oneIntegerArray = [6]
+    @oddNumbersArray = [1,3,5,7,1321]
```
pirvudoru: What changes do you need to do to add a new test following the convention of initializing all the test cases at once ?

tests/problem1Tests.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def teardown
+
+  end
```
pirvudoru: Are you required to implement this method in the test case even if it does nothing ?

tests/problem1Tests.rb
```diff
@@ -0,0 +1,37 @@
+require 'test/unit'
+require '../src/problem1'
+
+class Problem1Tests < Test::Unit::TestCase
+
+  def setup
+    @problem1 = Problem1.new
+    @randomIntegerTestData = [1,4,7,4,10,6]
+    @orderedIntegerTestData = (1..100).to_a
```
pirvudoru: I suggest following a coding convention for any programming language.

Ruby: https://github.com/bbatsov/ruby-style-guide

src/problem1.rb
pirvudoru, line 3: :+1:

tests/problem1Tests.rb
intellinor: Of course not, you are right, it can be scrapped :)

src/problem1.rb
intellinor: I used reduce because I read on ruby-doc that it applies a binary operation on all the elements of the array. Also in standard ruby language, in order to write as simple and clean as possible, I only saw this function, or inject. Not really sure if sum is present here, or just in rails framework.

alexandru-calinoiu: Got it,

No need to second guess we can always check the docs http://ruby-doc.org/core-2.4.2/Array.html#method-i-sum

Apparently `sum` is a little better when summing up float numbers, not the case in our problem https://stackoverflow.com/a/39933431/261363

alexandru-calinoiu:
Thank you, will reply via email.

PR #23, opened 2017-11-18
Feature branch - Problem2

intellinor:
Aici am complicat putin treaba intentionat, se putea mult mai simplu, dar voiam sa scot cat mai putine iteratii, cu acest minunat limbaj #fun :smiley:


alexandru-calinoiu (changes requested):
You missed the point of TDD, test driven development in a nutshell means you have to run through the r-g-refactor loop many times in short iterations.

So for this problem for example a chain of commits will be:

- (red) Add test for empty array
- (green) Empty array returns no maximums
- (red) Add test for array with one element
- (green) Return both maximus as the only element in the array
- (refactor) Rename calculate method

src/problem2.rb
```diff
@@ -0,0 +1,46 @@
...
+    compareValues maxValues,firstValue
+    if secondValue != nil
+      compareValues maxValues,secondValue
+    end
+  end
+
+  def compareValues(maxValues, newValue)
+
+    if maxValues[0] <= newValue || maxValues[1] <= newValue
+      replaceMaxNumbers maxValues, newValue
```
pirvudoru: `compareValues` does much more than it advertises (would expect it to return a comparison result aka one of [-1, 0, 1]). In this case it mutates one of the parameters.

src/problem2.rb
```diff
@@ -0,0 +1,46 @@
+class Problem2
+  def calculateMaximum(values) #theta(n/2) -> O(n)
```
pirvudoru: A good practice for comments (unless used for documentation) is to denote `why`, not `what`.

src/problem2.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def setMaxNumbers(maxValues, firstValue, secondValue)
+    if maxValues[0] == nil && maxValues[1] == nil
+      maxValues[0] = firstValue
+      maxValues[1] = secondValue.nil? ? firstValue : secondValue
+      return
+    end
+    compareValues maxValues,firstValue
+    if secondValue != nil
+      compareValues maxValues,secondValue
```
pirvudoru: Can you find the cause for calling `compareValues` twice and the `nil` checks ?

tests/problem2Tests.rb
```diff
@@ -0,0 +1,53 @@
...
+  def test_equalMaxValueHasExpectedResult
+    assert_equal([1111,1111],@problem2.calculateMaximum(@testEqualMaxValueData))
+  end
+
+  def test_emptyArrayHasExpectedResult
+    assert_equal([nil,nil],@problem2.calculateMaximum(@testDataEmptyArray))
+  end
+
+  def test_oneElementArrayHasExpectedResult
+    assert_equal([9,9],@problem2.calculateMaximum(@testDataHasOneElement))
```
pirvudoru: Don't believe that this is correct. You only have 1 number in the array.

pirvudoru (changes requested):

tests/problem2Tests.rb
intellinor: In the previous implementation, I was returning [9,null].
Then I read this comment:

(red) Add test for array with one element
(green) Return both maximus as the only element in the array

pirvudoru: :+1: Talked with @alexandru-calinoiu and any solution is ok since the problem statement can be interpreted in multiple ways. No change required here.

src/problem2.rb
intellinor: It's because i take numbers two by two, and the null check is required when we encounter arrays of odd length. This whole solution could have been much simpler if i would have taken numbers one by one.

src/problem2.rb
intellinor: evaluateValues or a similar name would be more clear maybe.

alexandru-calinoiu (changes requested):
Is getting better,

But is still very far of, you are using a lot of procedural code with impure functions.

If you like to go down this rode and break this problem into smaller functions than readup https://en.wikipedia.org/wiki/Pure_function and make all your functions pure.

PR #24, opened 2017-11-18
Feature branch - Problem3

intellinor:


src/problem3.rb
```diff
@@ -0,0 +1,33 @@
+class Problem3
+  def validateAndGetDuplicateValues(values)
+    @response = []
+    if(!values.nil? && (values.instance_of?(Array)))
+      if(values.all? {|i| i.is_a?(Integer)})
```
alexandru-calinoiu: The problem states `Try to find a solution by executing just one loop.`

What do you think the implementation of `all?` does? Hint: when you click on the method header in the ruby docs it will show you the c implementation.

alexandru-calinoiu (changes requested):
Does not respect problem constrains.

src/problem3.rb
intellinor: Well the solution is actually done in only one loop, that's just a pre-validation, probably not needed if we actually always provide integers. getDuplicateValues is the function that does all the work.

src/problem3.rb
```diff
@@ -0,0 +1,33 @@
...
+        if(values.all? {|i| i <= values.length})
+          getDuplicateValues(values, @response)
+          return filterDistinctResponse(@response)
+        else
```
alexandru-calinoiu: There are 2 loops here also.

get has an `each`
filter has a `uniq` (more or less a loop depending on the implementation)

The point is that you can solve this in one loop.

Your implementation is actually close to the solution, but is very hard to read with all the extra chars, you don't need the many parentheses (if, method calls) you added and usually blocks on multiple lines are written with [do](https://stackoverflow.com/questions/2122380/using-do-block-vs-brackets#2122457)

src/problem3.rb
```diff
@@ -0,0 +1,17 @@
+class Problem3
+  def getDuplicatedValues(values)
+    response = []
+    hash = Hash.new(-1)
```
alexandru-calinoiu: Still not there,

```
Do not sort the array or use bit vectors. Try to work within the array without using any other temporary data structures
```

Hash can qualify both as a bit vector and as a temporary data structure.

alexandru-calinoiu (changes requested):
Not yet the right solution.

src/problem3.rb
```diff
@@ -0,0 +1,23 @@
...
+      else
+        if values[value.abs()-1] > 0
+          values[value.abs()-1] = values[value.abs()-1] * -1
+        elsif values[value.abs()-1] > -values.count
+          response.push value.abs()
+          values[value.abs()-1] = values[value.abs()-1] - values.count
+        end
+      end
+    end
+    return response
```
alexandru-calinoiu: Still not right,

A couple of things:

- `Given an array of numbers having size N in which every number is between 1 and N, determine if there are any duplicates in it` so you are suppose to return a boolean not a array. The key to every good solution is understanding the problem.

- There a couple of foul code smells about your solution, 'Conditional Complexity',  'Oddball Solution' read more about it here https://blog.codinghorror.com/code-smells/

PS: you don't need a `return` in ruby.

alexandru-calinoiu (changes requested):

src/problem3.rb
intellinor: Really? Then it's easy, I was just refactoring on how to return the values, so that i will return unique values for the duplicates. Damn. Will clean it up.

tests/problem3Tests.rb
```diff
@@ -0,0 +1,44 @@
...
+    @testRandomMultipleOccurrenceDuplicatesData = [5,9,5,8,7,4,3,7,1,4,7,5]
+    @testSameValuesData = [1,1,1,1,1,1,1,1,1]
+    @testDataEmptyArray = []
+  end
+
+  def teardown
+
+  end
+
+  def test_inputDistinctValuesHasExpectedResult
```
alexandru-calinoiu: Your tests should be very easy to read and show the whole truth.

They basically need to broken down into 3 sections:

- prepare
- act
- assert

This sections can be even separated by a blank line.

Here is an example that I am know you will like https://msdn.microsoft.com/en-us/library/hh694602.aspx#Anchor_3

alexandru-calinoiu (changes requested):
Finally, problem solved. And your ruby looks almost decent 🥇 

Now let's fix the tests, please see my comment.

PR #25, opened 2018-04-17
Added Simple Test 1

RalucaMihaela:
I added the solving of the Simple Test 1 and 2.


drqCode (commented):
Why are all the files included in the repository like node modules? You should use .gitignore: https://www.atlassian.com/git/tutorials/saving-changes/gitignore.

RalucaMihaela:
I included all the files, because on some of my projects I had the situation where my app broke because of bad dependencies. Maybe in time the packages will change and if the app depends on npm module it can't be updated anymore.
	I think that this depends on the type of the project. I will use .gitignore to remain only the necessary files.

drqCode:
node_modules don't need to be included. In case of installing wrong dependencies with npm, there is package.lock file (it has to be added to git). See https://docs.npmjs.com/files/package-lock.json. :+1:

alexandru-calinoiu:
@RalucaMihaela the proper way to handle node modules is to use a `package.json` and the latest version of `npm` will generate a `package-lock.json` that will make sure the exact version of packages is installed on all machines.

We can ignore this issue for now.

alexandru-calinoiu:
For your consideration here are a couple of general guidelines:

- use js conventions, for example file names by conventions use - not camel case, https://stackoverflow.com/questions/7273316/what-is-the-javascript-filename-naming-convention
- don't 'overthink' simple solutions, pb1 states `given an array of integers` there is no reason for you to guard against floats or any other type of character.

RalucaMihaela:
I understand. I will use js naming convensions.  I tested for floats or other type of characters, because I thought that if I have an array of integers I have to make sure there are only integers elements.

duplicates-in-array.js
```diff
@@ -0,0 +1,20 @@
+module.exports = function (array, n){
+  var hasDuplicates = false;
+  var sum = 0;
+  var length = array.length;
+    if ( array.length === 0 ) return false;
+  	if( n != length ) return false;
```
alexandru-calinoiu, line 6: The code indentation here makes it really hard to read.

duplicates-test.js
```diff
@@ -0,0 +1,44 @@
...
+      it('If there are duplicates return true', function(){
+          var array = [1,4,5,2,4];
+          var n = 5;
+          assert.equal(app(array,n),true);
+      });
+      it('If there are not duplicates return false', function(){
+          var array = [1,4,5,2,3];
+          var n = 5;
+          assert.equal(app(array,n),false);
+      });
```
alexandru-calinoiu, line 41: How about a test for `[2, 2, 2]` ?

simpleTest2/appSecondPb.js
```diff
@@ -0,0 +1,22 @@
+module.exports = function (array) {
+
+  	var firstMAX = Number.NEGATIVE_INFINITY;
+    var secondMAX = Number.NEGATIVE_INFINITY;
+
+    if(array.length === 0) return 0;
+
+    for(var index = 0; index < array.length; index++) {
+            if(array.length === 1) return [array[index],0];
```
alexandru-calinoiu, line 9: This can be pulled up.

simpleTest1/appFirstPb.js
```diff
@@ -0,0 +1,24 @@
+module.exports = function (array) {
+
+    var sum = 0;
+    let notNumber = 0;
+
+    if(array.length > 0) {
```
alexandru-calinoiu, line 6: Do you really need this here, what will happen if you don't have it and the array passed is `[]`?

alexandru-calinoiu (changes requested):
Apart from the inline comments related to problem 3:

- the form of the code is bad, code indentation is missing and it looks sloppy
- added to many files to the repo
 
On the good side:

- the tests are good and testing as many cases as possible.

duplicates-in-array.js
RalucaMihaela: Yes, I see. Usually I am more careful with this. In my files  those two conditions are aligned.

duplicates-test.js
RalucaMihaela: Yes, for this test my code returns that there are no duplicates. I tested for situations where I have numbers from 1 to n.  In this case I can verify if there are duplicates using another array. In this array I will store the original array and I will compare each item with the original array. I do not know how to do this only with one loop and without another structure.

duplicates-in-array.js
```diff
@@ -0,0 +1,20 @@
+module.exports = function (array, n){
```
drqCode, line 1: Why do you need `n` as parameter? The problem description says: "Given an array of numbers having size N in which every number is between 1 and N". In my opinion, it doesn't state that you'll be given an array of numbers and a number representing the size or the number of elements.

simpleTest2/appSecondPb.js
RalucaMihaela: Now I see that I do not need this as I am already printing [1,-Infinity] if I have just one item like [1].

simpleTest1/appFirstPb.js
RalucaMihaela: If I don't have it, the sum will be 0, so yes I do not need that condition there.

duplicates-in-array.js
RalucaMihaela: Yes it does not say, but I used n to verify if my given array has the number of items I want and to verify if each item is smaller than n.  I could have used only the array's length to verify if each number is smaller than the length.

drqCode: Given the description, you do not need to check if the array is a good one (that's others responsibility), you only have to check for duplicates. :smiley:

alexandru-calinoiu:
Closing this as it appears it has been replace by https://github.com/Agilefreaks/Test/pull/26

duplicates-test.js
alexandru-calinoiu: Not sure what you mean, [2,2,2] is between 1..n

There is a way to solve this problem with only one datastructure, notice that you can update/destroy the existing array.

PR #26, opened 2018-04-24
Review changes

RalucaMihaela:


alexandru-calinoiu:
There is no point opening a new pr you can simply reply to the existing  pr or comment.
(no need to change this one now as  I closed the previous pr, will continue or review here)

levelboy:
Why did you use two different naming conventions? (e.g. someFile.js and some-file.js)

simpleTest1/appFirstPb.js
```diff
@@ -0,0 +1,19 @@
+module.exports = function (array) {
+    
+    var sum = 0;
+    let notNumber = 0;
+
+    for (var index = 0; index < array.length; index++) {
+        if (parseInt(array[index]) === array[index]) {
```
levelboy: Is there the possibility for items in the array to not be numbers?

simpleTest1/appFirstPb.js
```diff
@@ -0,0 +1,19 @@
...
+            if (array[index] % 2 === 0) {
+                sum += array[index];
+            }
+        } else {
+            notNumber++;
+        }
+    }
+    if (notNumber === array.length) {
```
levelboy: Is this part of the problem statement?

simpleTest1/test/appFirstPbTest.js
```diff
@@ -0,0 +1,52 @@
...
+      });
+      it('Just even elements should return sum of element', function(){
+        var array = [2, 2, 2];
+        assert.equal(app(array),6);
+      });
+      it('Float and integer elements should return the sum of even integers', function(){
+        var array = [1.2, 2, 3];
+        assert.equal(app(array),2);
+      });
+      it('Even and odd elements should return sum of even elements', function(){
```
levelboy, line 43: This description is not what the setup does. The setup adds only even numbers.

simpleTest2/appSecondPb.js
```diff
@@ -0,0 +1,22 @@
...
+    for(var index = 0; index < array.length; index++) {
+            if(parseInt(array[index]) === array[index]) {
+                  if(array[index] > firstMAX ) {
+                          secondMAX  = firstMAX ;
+                          firstMAX  = array[index];
+                  } else if (array[index] > secondMAX ) {
+                        	secondMAX  = array[index];
+                  }
+            } else { 	 
+                  secondMAX = 0;
```
levelboy: Why not set them back to `Number.NEGATIVE_INFINITY` ?

simpleTest2/appSecondPb.js
```diff
@@ -0,0 +1,22 @@
+module.exports = function (array) {
+
+  	var firstMAX = Number.NEGATIVE_INFINITY;
+    var secondMAX = Number.NEGATIVE_INFINITY;
+
+    if(array.length === 0) return 0;
+
+    for(var index = 0; index < array.length; index++) {
+            if(parseInt(array[index]) === array[index]) {
```
levelboy: As with the first problem, it would be safe to assume that all that you receive are integers

duplicates-in-array.js
```diff
@@ -0,0 +1,23 @@
+module.exports = function (array, n) {
+
+    var length = array.length;
+    var counter = zeroFilledArray(array.length);
+
+    if (array.length === 0) return false;
```
levelboy: You already have a variable for `array.length`

duplicates-in-array.js
```diff
@@ -0,0 +1,23 @@
+module.exports = function (array, n) {
+
+    var length = array.length;
+    var counter = zeroFilledArray(array.length);
```
levelboy: You already have a variable for `array.length`

duplicates-in-array.js
```diff
@@ -0,0 +1,23 @@
...
+
+    if (array.length === 0) return false;
+    if (n != length) return false;
+    for (var index = 0; index < array.length; index++) {
```
levelboy: You already have a variable for array.length

duplicates-in-array.js
```diff
@@ -0,0 +1,23 @@
+module.exports = function (array, n) {
```
levelboy: Does the problem statement mention that you receive two parameters for this function?

duplicates-in-array.js
```diff
@@ -0,0 +1,23 @@
...
+    var length = array.length;
+    var counter = zeroFilledArray(array.length);
+
+    if (array.length === 0) return false;
+    if (n != length) return false;
+    for (var index = 0; index < array.length; index++) {
+      if (array[index] < 1 || array[index] > n) {
+          return false;
+      } else {
+    		if(counter[array[index]] === 1){
```
levelboy: Can you think of a method to not use the bit array to verify for duplicates?

RalucaMihaela:
> Why did you use two different naming conventions? (e.g. someFile.js and some-file.js)

After I did the first two problems, Alex recommended me to use javascript naming convention so I tried to use it only on the third problem, because I added it the last one. I will change the naming convetion for the others too.

simpleTest1/appFirstPb.js
RalucaMihaela: The problem said that I have to do the sum of even numbers. I know that even/odd only refers to integers, so that is why I verified If I had only integers. An array can have different types of items, of string type, boolean type etc. On best-case scenario I think I can assume I have only integers.

simpleTest1/test/appFirstPbTest.js
RalucaMihaela: Yes I see. My mistake, I did not see that I did not change the test array.

simpleTest2/appSecondPb.js
RalucaMihaela: Yes, in this part best way is to set the variables with the values with which they were are initialized, but I didn't like to see -INFINITY, that is why I assigned those two variables with 0.

simpleTest2/appSecondPb.js
RalucaMihaela: Ok, I understand

duplicates-in-array.js
RalucaMihaela: No it does not say. As I said to Costel on his review, I used n to verify if my given array has the number of items I want and to verify if each item is smaller than n. I could have used only the array's length to verify if each number is smaller than the length, because in the description does not say that I have to check if the array is in correct form.

duplicates-in-array.js
RalucaMihaela: At first I did this problem by verifying if the sum of all numbers is n(n+1)/2.  If it wasn't equal then I had duplicates, but then I saw that this case didn't work If I did not have the array as a series from 1 to n.  I will try to do it in another way.

duplicates-test.js
```diff
@@ -0,0 +1,49 @@
+const assert = require('assert');
+const app = require('../duplicates-in-array');
+
+describe('App', function() {
```
levelboy, line 4: Tests don't seem to reflect the last modifications to the function since https://github.com/Agilefreaks/Test/pull/26/files#diff-9b7a88c0978b5162b6f6b6e273897839R1 has one parameter but you still call it with two.

duplicates-in-array.js
```diff
@@ -0,0 +1,14 @@
+module.exports = function (array){
+  for(var index = 0; index < array.length; index++){
+       var item = array[Math.abs(array[index])];
```
levelboy, line 3: `array[Math.abs(array[index])]` shows up in three places in this method

duplicates-in-array.js
```diff
@@ -0,0 +1,14 @@
...
+       if (array.length === 0) return false;
```
levelboy, line 4: Why is this `if` in the for loop?

duplicates-in-array.js
```diff
@@ -0,0 +1,14 @@
+module.exports = function (array){
+  for(var index = 0; index < array.length; index++){
```
levelboy, line 2: You can iterate over the array with `forEach()`

levelboy (commented):
Indentation is all messed:
There are:
- 8 spaces
- 4 spaces
- 2 spaces
- 1 tab

Choose one and stick to it

levelboy:
Closing this. Thanks!

PR #27, opened 2018-04-30
First problem

HiciuStefan:


first_problem.rb
```diff
@@ -0,0 +1,23 @@
+class FirstProblem
+
+  def initialize(input_array)
+    @numbers_array = input_array
+  end
+
+  def compute()
+    if @numbers_array.empty?
```
levelboy: Do you need this check here? If you delete this, and rerun the tests, they should all pass.

first_problem_test.rb
```diff
@@ -0,0 +1,39 @@
+require 'test/unit'
+require_relative'./first_problem'
+
+class FirstProblemTest < Test::Unit::TestCase
+
+  # Called before every test method runs. Can be used
+  # to set up fixture information.
+  def setup
+    # Do nothing
```
levelboy: There's no need to have methods that do nothing

first_problem_test.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def test_compute_emptyArray_0
+    array = Array.new
+    assert_equal(0,FirstProblem.new(array).compute)
+  end
+
+  def test_compute_noEvenNumbersArray_0
+    array = [1]
+    assert_equal(0,FirstProblem.new(array).compute)
```
levelboy: You should have a separate unit test for each of the asserts here

first_problem_test.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    array = [1,3,5,7]
+    assert_equal(0,FirstProblem.new(array).compute)
+  end
+
+  def test_compute_validNumbersArray_correctResult
+    array = [1,3,8]
+    assert_equal(8,FirstProblem.new(array).compute)
```
levelboy: You should have a separate unit test for each of the asserts here

first_problem.rb
```diff
@@ -0,0 +1,23 @@
...
+    then
+      return 0
+    end
+
+    sum = 0
+    @numbers_array.each do |number|
+      if number % 2 == 0
```
levelboy: Ruby is quite a nice language. Given that numbers are objects, they respond to '.even?' or '.odd?'
Can you write this first problem in one line?

levelboy (changes requested):

first_problem_test.rb
HiciuStefan: I would argue the same as before , since we respect the expected input , there should be no need for multiple unit tests

first_problem_test.rb
HiciuStefan: Why ? The asserts are against the expected input. 
Both the arrays have no even numbers

first_problem_test.rb
HiciuStefan: On the IDE i used it generates test classes with these methods .  
As i don't have plenty ruby experience i assumed that these methods are ok to remain there. However i would normally not have empty methods . 
I will remove them

first_problem.rb
HiciuStefan: I can sure try ! :)

first_problem.rb
HiciuStefan: TDD tryouts ^^ 
I did unit test for empty array first , but yes now that i look over it again there is no problem to remove the check.

first_problem.rb
HiciuStefan: not sure the result is what you expected

first_problem_test.rb
HiciuStefan: i took liberty to remove these methods from the other branches also

first_problem.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def initialize(input_array)
+    @numbers_array = input_array
+
+  end
+
+  def compute()
+    sum = 0
+    @numbers_array.each do |number|
+      sum+= number if number.even?
```
alexandru-calinoiu: Ok this one option on how to do it, but ruby is really more expressive (as you've noticed with `even?`). Can you figure out how to do this in 1 line?

first_problem_test.rb
alexandru-calinoiu: There a couple of reasons why having multiple asserts in a tests is not a good idea. The most obvious one is that you where asked to do tdd, that means write a test that fails (and it will fail after the first assert), fix it (the minimum code required to make the test pass) and then refactor.

alexandru-calinoiu (approved):
Is good enough, but it can be so much better :)

first_problem_test.rb
HiciuStefan: I guess i didn't quite understand that you wanted me to do 1 failing unit test , 1 code change , 1 refactor (if neccesary) at a time.
I did more of a add unit tests (failing) for functionality , do a "demo" version for 1st one passing then 2/3rd failing . 
If i was developing TDD i would imagine i don't commit every time i have created 1 failing unit test. To then commit again with it working. To then commit again when refactor. At least in my mind i would commit when it's done once.
But this is just how i envisioned real working scenario. I could be totally wrong. And that's why all my PR's look like that.

Related to having one assert per unit test , in this particular case , i could have just merged them. There is no correlation between size of array and output.
I agree that maybe it could be confusing to have 2 assert for the same validation.

first_problem.rb
HiciuStefan: i guess i might have done it !

PR #28, opened 2018-04-30
Third problem

HiciuStefan:


third_problem.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      if pointingTowardsItself(currentNumber, it)
+        @numbers[it] = 0
+        next
+      end
+
+      targetIndex = currentNumber - 1
+      currentIndex = it
+
+      while !duplicate
```
alexandru-calinoiu, line 27: If you read the problem statement carefully it requires not to use nested loops, you are not doing this.

alexandru-calinoiu (changes requested):
Does not respect all the conditions in the problem statement.

third_problem.rb
```diff
@@ -1,13 +1,68 @@
...
   end
 
   def hasDuplicates
-    if @numbers.empty?||@numbers.length==1 then
+    if @numbers.empty? || @numbers.length == 1 then
       return false
     end
+    duplicate = false
+
+    for it in 0...@numbers.length
```
pirvudoru: Just a thing to remember:
We write code for humans, not machines. It is very difficult to follow the code if the naming is not self-descriptive. In this case one has to do the association `it` = `index`

third_problem.rb
```diff
@@ -0,0 +1,68 @@
+class ThirdProblem
+  def initialize(numbers_array)
+    @numbers = numbers_array
```
pirvudoru: Mixed convention. You use `camelCase` as well as `snake_case`. You should be sticking to one.

pirvudoru (changes requested):
The solution is incorrect:

image

third_problem.rb
HiciuStefan: fixed

third_problem.rb
HiciuStefan: Actually maybe i read it too carefully
"Do not use 2 nesting for loops since the time complexity would be high".
I thought that the reason to not use 2 loops is because of the time complexity. My solution does not increase in time complexity. It's actually O(n) still. In worst case it's O(2n) which is still O(n).
That's why i considered i respected all the conditions.
If you just don't want to have loop in loop than i guess i failed

third_problem.rb
pirvudoru: https://github.com/bbatsov/ruby-style-guide

third_problem.rb
HiciuStefan: For an iterator between 0 and x I would consider "it" as reasonable enough for a simple use-case like this . However when code gets more complex i agree it's worth to put a better name.
Would you prefer that i change this to index ? I personally don't see that much the added value.

third_problem.rb
HiciuStefan: I know , i should have just used the something_something everywhere. Personally i find it very unappealing and that's the reason why i preferred to stick to my camelCase .
Of course , working on a daily basis in Ruby i would force myself to respect those conventions.

HiciuStefan: also, updated the code to have correct solution.

third_problem.rb
alexandru-calinoiu: Indeed, but in can be solved only with one loop.

PR #29, opened 2018-04-30
Second problem

HiciuStefan:


alexandru-calinoiu:
@HiciuStefan the requirement in the email was to solve this problems using TDD, there is a link in the email on what that means, to sum it up:

Step1: write a failing test (commit), red
Step2: write just enough code to make test pass (commit), green
Step3: refactor, if the case might be (commit)

This is not visible from your commits and tests

second_problem.rb
```diff
@@ -0,0 +1,32 @@
+class SecondProblem
+
+  attr_reader :never_ever_used_value
+
+  def initialize(numbers_array)
+    @numbers = numbers_array
+  end
+
+  def getBiggestTwoNumbers
+    minValue = -1234567890
```
alexandru-calinoiu: How did you come up with this number?

min_max.rb
```diff
@@ -0,0 +1,8 @@
+class MinMax
+  attr_reader :biggerMaxVal, :smallerMaxVal
+
+  def initialize(smallerMaxVal, biggerMaxVal)
```
alexandru-calinoiu: Having a class for this is not very efficient, have a look at [structs](https://ruby-doc.org/core-2.5.0/Struct.html)

second_problem.rb
```diff
@@ -0,0 +1,32 @@
...
+        secondLarge = number
+        if secondLarge > firstLarge
+          temp = firstLarge
+          firstLarge = secondLarge
+          secondLarge = temp
+        end
+      end
+    end
+
+    return MinMax.new(secondLarge, firstLarge)
```
alexandru-calinoiu: Interesting solution, but what happens for the following array [4, 4, 3, 2]? Can you argue that [4. 3] is not the correct solution?

second_problem.rb
HiciuStefan: my version of MAXINT since it doesn't have one. didn't want to bloat the solution with creating it

second_problem.rb
HiciuStefan: My solution would result in 4 ,4 
image

HiciuStefan: added a unit test to satisfy this situation

min_max.rb
HiciuStefan: done !

PR #30, opened 2018-05-03
First problem

HiciuStefan:


first_problem_test.rb
```diff
@@ -0,0 +1,11 @@
+require 'test/unit'
+require_relative'./first_problem'
+
+class FirstProblemTest < Test::Unit::TestCase
+
+  def test_compute_empty_array_0
+    array = Array.new
+    assert_equal(-1,FirstProblem.new(array).compute)
```
alexandru-calinoiu: Not very good at number theory but shouldn't this be `0`?

HiciuStefan: i was trying to push for a "refactor" kind of step. Seems rather weird ideea now. Don't know exactly what i was thinking
But yes , it's 0...

first_problem.rb
```diff
@@ -0,0 +1,16 @@
...
+  def initialize(input_array)
+    @numbers_array = input_array
+  end
+
+  def compute()
+    sum = 0
+    @numbers_array.each do |number|
+      sum += number if number.even?
+    end
+    sum
```
alexandru-calinoiu: The `sum` lines in this code look really strange, are you familiar with the `Enumerator` concept? Can you express this as a concatenation of operations?

first_problem_test.rb
```diff
@@ -0,0 +1,21 @@
+require 'test/unit'
+require_relative'./first_problem'
+
+class FirstProblemTest < Test::Unit::TestCase
+
+  def test_compute_empty_array_0
+    array = Array.new
```
alexandru-calinoiu, line 7: Usually unit tests have 3 regions: `arrange`, `execute` and `assert`, is usually a good idea to have those show in the test.

(nothing to do for now just sharing insights)

first_problem.rb
HiciuStefan: updated code

first_problem.rb
```diff
@@ -0,0 +1,14 @@
+class FirstProblem
+
+
+  def initialize(input_array)
+    @numbers_array = input_array
+  end
+
+  def compute()
+    sum = 0
+    @numbers_array.each {|number| sum+=number if number.even? }
```
alexandru-calinoiu, line 10: I would have go with something like:

```
numbers.select(&:even?).inject(:+)
```

alexandru-calinoiu:

PR #31, opened 2018-05-03
Second problem

HiciuStefan:


second_problem.rb
```diff
@@ -0,0 +1,33 @@
+class SecondProblem
+
+  attr_reader :min_value
+
+  def initialize(numbers_array)
+    @numbers_array = numbers_array
+    Struct.new("MinMax", :second_large, :first_large)
+    @min_value = -1
```
alexandru-calinoiu: I don't think this is correct, integers have negative numbers smaller than -1, so for example an array of [-2, -3, -4] should return [-2, -3].

In the initial solution you have a large negative number, I wanted to know how you arrived at that number.

alexandru-calinoiu (changes requested):
Please review my comment.

second_problem.rb
HiciuStefan: My ideea was to get a "big" number to have a kind of -MAXINT. That's why the -1234567890.
In this iteration i just considered positive numbers. -> bad
I've updated the code to have this -MAXINT value.

second_problem.rb
```diff
@@ -0,0 +1,38 @@
+class SecondProblem
+
+  N_BYTES = [42].pack('i').size
+  N_BITS = N_BYTES * 16
+  MAX = 2 ** (N_BITS - 2) - 1
```
alexandru-calinoiu, line 5: This makes sens :+1:

alexandru-calinoiu:

PR #32, opened 2018-05-03
Third problem

HiciuStefan:


third_problem.rb
```diff
@@ -0,0 +1,22 @@
+class ThirdProblem
+  def initialize(numbers_array)
+    @numbers = numbers_array
+  end
+
+  def hasDuplicates
```
alexandru-calinoiu, line 6: Much better :+1:

third_problem.rb
```diff
@@ -0,0 +1,22 @@
...
+    found = false
+    @numbers.each {|index|
+      absolute_current_index =index.abs-1
+      absolut_target_index = @numbers[absolute_current_index].abs-1
```
alexandru-calinoiu, line 10: You keep repeating `@numbers[absolute_current_index]`, I will think is worth extracting in a variable. Index lookup for arrays in ruby has it's cost.

third_problem.rb
```diff
@@ -0,0 +1,22 @@
...
+      if @numbers[absolut_target_index] < 0
+        found = true
+        break
+      end
+      @numbers[absolut_target_index] = -@numbers[absolut_target_index]
+    }
+    found
```
alexandru-calinoiu: If you where allowed to use a different data structure to solve this problem what data structure will you use and why?

third_problem.rb
HiciuStefan: sure , extracted to variable

alexandru-calinoiu:

PR #34, opened 2018-05-04
Test

anametiu:


Problem1.rb
```diff
@@ -0,0 +1,18 @@
+def  sum_even_numbers(array)
+
+  sum = 0
+  array.each { |x| sum += x.even? ? x : 0 }
```
drqCode: Can you make it one line (without any aux variable)?

Problem2.rb
```diff
@@ -0,0 +1,32 @@
...
+end
+
+describe 'Problem2' do
+
+  it 'should find max numbers' do
+
+    puts expect(find_max_numbers([4, 8, 2, 6, 9, 3])).to eq([9, 8])
+    puts expect(find_max_numbers([1, 2, 3, 4, 5, 6])).to eq([6, 5])
+    puts expect(find_max_numbers([3, 2, 7, 1, 1, 1])).to eq([7, 3])
+
```
drqCode: What if the array contains negative numbers?

Problem3.rb
```diff
@@ -0,0 +1,48 @@
...
+def contains_duplicates?(array)
+
+  min = 1
+
+  array.each_index do |i|
+
+    temp = array[i]
+    j = temp - min
+
+    while i != j do
```
drqCode: "Do not use 2 nesting for loops since the time complexity would be high. Try to find a solution by executing just one loop". Consider using just one loop, not matter if there is a very small one. :)

Problem2.rb
alexandru-calinoiu: Indeed integer numbers can be negative.

alexandru-calinoiu (changes requested):

Problem3.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  return false
+
+end
+
+describe 'Problem3' do
+
+  it 'should find duplicates' do
+
+    puts contains_duplicates?([1, 2, 3, 4, 2])
```
alexandru-calinoiu: Tests should contain asserts

alexandru-calinoiu (changes requested):
Very good try, but problem 3 does not respect the requirements, you are suppose to solve it with only one loop.

Problem3.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      j = temp - min
+
+    end
+
+  end
+
+  return false
+
+end
```
pirvudoru: Why all the empty lines?

Problem3.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  array.each_index do |i|
+
+    temp = array[i]
+    j = temp - min
+
+    while i != j do
+
+      if temp == array[j]
+        return true
```
pirvudoru: Please keep a single return per function. Ruby, to help with this, returns the result of the last line in the method, without needing to specifically write `return`. Ex:

```
def sum(a,b)
  a + b
end

sum(1,3) # returns 4
```

Problem2.rb
```diff
@@ -0,0 +1,32 @@
...
+end
+
+describe 'Problem2' do
+
+  it 'should find max numbers' do
+
+    puts expect(find_max_numbers([4, 8, 2, 6, 9, 3])).to eq([9, 8])
+    puts expect(find_max_numbers([1, 2, 3, 4, 5, 6])).to eq([6, 5])
+    puts expect(find_max_numbers([3, 2, 7, 1, 1, 1])).to eq([7, 3])
```
pirvudoru: Why do you have "puts" for every expect ?

Problem2.rb
```diff
@@ -0,0 +1,32 @@
...
+    if x > first
+      second = first
+      first = x
+    elsif x > second
+      second = x
+    end
+
+  end
+
+  return Array.new([first, second])
```
pirvudoru: Just for reference: https://github.com/bbatsov/ruby-style-guide
Every major programming language has it's own.

Problem3.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+
+     swap(array, i, j)
+
+      j = temp - min
+
+    end
```
pirvudoru: We write code for people, not machines. Spent ~ 15 minutes to figure out what the code does, but still a mistery to me. Can you please describe what you wanted to do ?

Btw, it is not 100% working

image

pirvudoru (changes requested):

pirvudoru:
Also, can check out: http://blog.cleancoder.com/uncle-bob/2014/12/17/TheCyclesOfTDD.html

Problem3.rb
anametiu: I uploaded tests for problem1 and problem2 using assert

Problem3.rb
anametiu: The algorithm iterates over the array and moves each number into its corresponding index position. If the number is already found there, it means it's a duplicate. Indeed algorithm fails for array[3, 4, 1, 2]. I will correct it.

anametiu: I completly reimplemented problem3 to meet the O(n) requirements

alexandru-calinoiu:
@anametiu is really hard to review the code changes, please have only one solution for each problem. right now there are 2 files 'problem3' and 'Problem3'

alexandru-calinoiu:
Closing this due to lack of activity.

alexandru-calinoiu:
@anametiu I've reopened this following the email I've received from you.

Problem3.rb
anametiu: I checked again the implementation and you're right. Also the method swap extracted for refactoring makes the algorithm to fail for certain tests. I completely re implemented the requirement and committed in 79f5fe7284839366e7dcbe32786b4a2bb38b3243

Problem3.rb
anametiu: I refactored all my code to take into consideration your remark

Problem3.rb
anametiu: I thought they will make the code easier to read. I removed them

Problem3.rb
anametiu: Algorithm fails because of the swap method.

Problem1.rb
anametiu: Yes: array.select(&:even?).sum
Commit: d2d8a7489a6358e37802db6ee26054408c72efad

Problem2.rb
anametiu: Indeed, after reading the guide I got rid of the empty lines and unnecessary returns in all problems

Problem2.rb
anametiu: Actually I looked more into testing and implemented unit tests in separate files using assert

Problem2.rb
anametiu: True, I initialized the first and second variables with minus infinite. Commit: 3664186d8d12b3b17bfeb43baaa175931c2f386b

Problem3.rb
anametiu: I implemented a different algorithm in commit 79f5fe7284839366e7dcbe32786b4a2bb38b3243

alexandru-calinoiu:
Closing this because it to old..

PR #35, opened 2018-05-19
Solutions

bunea:


src/Test.First/SumEvens.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+
+namespace Test.First
+{
+    public class SumEvens
+    {
+        public int Sum(int[] arr)
+        {
+            var sum = 0;
+            foreach (var number in arr)
```
alexandru-calinoiu: There is a cool extension to the c# language called linq that makes code ever more expressive.

Can you rewrite you re solution using linq?

src/Test.Second/LargestTwo.cs
```diff
@@ -0,0 +1,32 @@
...
+                    second = first;
+                    first = num;
+                }
+                else if (second < num)
+                {
+                    second = num;
+                }
+            }
+
+            return new int[] { first, second };
```
alexandru-calinoiu: Is there a better structure to store a result composed only of 2 values?

alexandru-calinoiu (commented):
Looks nice an clean, made a couple of comments just for reference.

src/Test.Second/LargestTwo.cs
bunea: A `ValueTuple`. I've refactored the method to return a `ValueTuple`. I didn't use it the first time around because xunit doesn't support one in `InlineData`

alexandru-calinoiu:
Nice clean solutions, closing this and will followup via email.

PR #36, opened 2018-06-23
Test solutions

crisan-catalin:
Hello,
This is the test solutions.


src/math_util.rb
```diff
@@ -0,0 +1,43 @@
+class MathUtil
+
+  def self.sum_even_numbers(numbers)
+    numbers&.select(&:even?)&.sum
```
levelboy: How can you remove the safe navigation operator `&.` from this?

src/math_util.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def self.two_largest_numbers(numbers)
+    return nil if numbers.nil? ||  numbers.size < 2
```
levelboy: Same as before, how can you remove the checks for `nil`?

levelboy: I imagine you'd use it like:
`firstNumber, secondNumber = MathUtil.two_largest_numbers(arrayOfNumbers)`

How do you use it when it returns `nil`?

levelboy: Given that you chose to do validations (even though it was not specifically asked for in the problem), why did you chose to return `nil` in case the input is invalid instead of raising an error?
This applies for all three problems

src/math_util.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+    }
+
+    [firstNumber, secondNumber]
+  end
+
+  def self.exists_duplicates(numbers)
+    return nil if numbers.nil? || numbers.size < 2
+
+    numbers.each { |x|
```
levelboy: Not a big issue, but when a block of code spans on more than one line, then it is better to use `do ... end` block

src/math_util.rb
```diff
@@ -0,0 +1,43 @@
...
+      elsif x > secondNumber
+        secondNumber = x
+      end
+    }
+
+    [firstNumber, secondNumber]
+  end
+
+  def self.exists_duplicates(numbers)
+    return nil if numbers.nil? || numbers.size < 2
```
levelboy: Given that you have to find if there are duplicates in an array or not, how does it help you that you return `nil`? Ask yourself how would you use a function like this? How can it help you in how you'll use it if it returns `nil`?

test/math_util_test.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  def test_two_largest_numbers_empty_array_returns_nil
+    assert_equal(nil, MathUtil.two_largest_numbers([]))
+  end
+
+  def test_two_largest_numbers_one_element_returns_nil
+    assert_equal(nil, MathUtil.two_largest_numbers([0]))
+  end
+
+  def test_two_largest_numbers_returns_4_and_3
```
levelboy: Usually when you describe a test, you should mention the function you are testing, the context that you are testing and how you expect it to behave:
This is good: https://github.com/Agilefreaks/Test/pull/36/files#diff-6cf82e7fbd372cdd4efcc55f35b88704R12
The line where this comment is attached to is not

src/math_util.rb
```diff
@@ -0,0 +1,43 @@
+class MathUtil
```
levelboy: What was the reasoning to put this in a MathUtil since this has operations on arrays of numbers?

levelboy (changes requested):
@crisan-catalin Please see and respond to comments

src/math_util.rb
crisan-catalin: I wasn't inspired, I think it should be named ArrayUtil.

src/math_util.rb
crisan-catalin: I use it like:
```
result = MathUtil.two_largest_numbers(arrayOfNumbers)
```
where ```result``` is an array.

Also, I read this post "<i>Avoiding “NoMethodError” for nil in Ruby</i>": https://dlet.me/avoiding-nomethoderror-for-nil-in-ruby-9ce0f3fd0e30

test/math_util_test.rb
crisan-catalin: I will refactor that code.

src/math_util.rb
crisan-catalin: I can raise an exception or return ```false```. But I think the best solution is to return ```false```.
What do you think?

src/math_util.rb
crisan-catalin: If I receive a ```nil``` parameter I need to return ```0``` because the sum is 0 or I can return ```nil``` or raise an exception.
What else can I do?

src/array_util.rb
```diff
@@ -0,0 +1,43 @@
+class ArrayUtil
+
+  def self.sum_even_numbers(numbers)
+    numbers&.select(&:even?)&.sum
```
alexandru-calinoiu: Why use the safe operator `&`?

Is there a better alternative?

src/array_util.rb
```diff
@@ -0,0 +1,43 @@
...
+    firstNumber = numbers[0]
+    secondNumber = numbers[1]
+
+    if firstNumber < secondNumber
+      firstNumber, secondNumber = secondNumber, firstNumber
+    end
+
+    numbers[2...numbers.size].each do |x|
+      if x > firstNumber
+        secondNumber = firstNumber
```
alexandru-calinoiu, line 19: Why did you choose to do the assignments separately here and initially you condensed them to one line?

What is the better approach in your opinion?

src/array_util.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    [firstNumber, secondNumber]
+  end
+
+  def self.exists_duplicates(numbers)
+    return nil if numbers.nil? || numbers.size < 2
+
+    numbers.each do |x|
+      if numbers[x.abs - 1].to_i < 0
```
alexandru-calinoiu: Why do you need `.to_i` ?

src/array_util.rb
```diff
@@ -0,0 +1,43 @@
+class ArrayUtil
+
+  def self.sum_even_numbers(numbers)
+    numbers&.select(&:even?)&.sum
+  end
+
+  def self.two_largest_numbers(numbers)
+    return nil if numbers.nil? ||  numbers.size < 2
```
alexandru-calinoiu: Are you sure this is correct? If the array has only 1 number returning `[n, nil]` is not correct?

src/array_util.rb
```diff
@@ -0,0 +1,43 @@
...
+        firstNumber = x
+      elsif x > secondNumber
+        secondNumber = x
+      end
+    end
+
+    [firstNumber, secondNumber]
+  end
+
+  def self.exists_duplicates(numbers)
```
alexandru-calinoiu, line 29: There is a lot of repetition for `x.abs - 1` consider refactoring.

src/array_util.rb
```diff
@@ -0,0 +1,43 @@
...
+    return nil if numbers.nil? || numbers.size < 2
+
+    numbers.each do |x|
+      if numbers[x.abs - 1].to_i < 0
+        return true
+      else
+        numbers[x.abs - 1] = numbers[x.abs - 1].to_i * (-1)
```
alexandru-calinoiu: Multiplying by -1 looks a little strange, what is you write `-numbers[index]`

src/array_util.rb
```diff
@@ -0,0 +1,43 @@
+class ArrayUtil
```
alexandru-calinoiu, line 1: Strange name for a class, why did you choose to have the methods in a class?

test/array_util_test.rb
```diff
@@ -0,0 +1,76 @@
+require_relative "../src/array_util"
+require "test/unit"
+
+class TestArrayUtil < Test::Unit::TestCase
+
+# Test sum_even_numbers
+
+  def test_sum_even_numbers_nil_value_returns_nil
+    assert_nil(ArrayUtil.sum_even_numbers(nil))
```
alexandru-calinoiu: Superfluous test, the problem statement clearly states that you are given an array, so you can safely expect an array.

Now that you have it you need to improve the implementation, see my comment.

test/array_util_test.rb
```diff
@@ -0,0 +1,76 @@
...
+  def test_two_largest_numbers_empty_array_returns_nil
+    assert_equal(nil, ArrayUtil.two_largest_numbers([]))
+  end
+
+  def test_two_largest_numbers_one_element_array_returns_nil
+    assert_equal(nil, ArrayUtil.two_largest_numbers([0]))
+  end
+
+  def test_two_largest_numbers_positive_numbers_array_returns_4_and_3
+    assert_equal([4, 3], ArrayUtil.two_largest_numbers([0, 1, 2, 3, 4]))
```
alexandru-calinoiu: The test is a little bias as it passes the arrays sorted.

If this where all the tests the implementation could have been simplified.

test/array_util_test.rb
```diff
@@ -0,0 +1,76 @@
...
+  def test_exists_duplicates_with_multiple_duplicates_is_true
+    assert_true(ArrayUtil.exists_duplicates([1, 2, 3, 1, 1, 1, 2, 3, 1]))
+  end
+
+  def test_exists_duplicates_without_duplicates_is_false
+    assert_false(ArrayUtil.exists_duplicates([1, 2, 3]))
+  end
+
+  def test_exists_duplicates_with_negative_duplicates_is_true
+    assert_true(ArrayUtil.exists_duplicates([-1, 2, 3, -1]))
```
alexandru-calinoiu: This is wrong, the problem clearly states that each number is between `1 and N`

In case you allow negative numbers your solution will fail `-1, 1, 2` will return true but there are no duplicates.

alexandru-calinoiu (changes requested):
Excellent start, but solutions need a little more improvement.

src/array_util.rb
crisan-catalin: I use safe operator to return ```nil``` if method is called with ```nil``` parameter.
I miss something?

src/array_util.rb
crisan-catalin: I don't need it. I was testing something and I forgot to remove.

src/array_util.rb
crisan-catalin: Done.

src/array_util.rb
crisan-catalin: Done

test/array_util_test.rb
crisan-catalin: That's true.

src/array_util.rb
crisan-catalin: Update:
I don't need to use safe operator or to check for ```nil``` because the problem says I have an array.
I thought I had to check for ```nil``` to be sure.

src/array_util.rb
crisan-catalin: I was reading about Ruby conventions: https://github.com/github/rubocop-github/blob/master/STYLEGUIDE.md
But I miss something: <i>"Prefer {...} over do...end for single-line blocks. Avoid using {...} for multi-line blocks (multiline chaining is always ugly)"</i>

So, this approach is better for multi-line.

src/array_util.rb
crisan-catalin: I tested ```max(n)``` method from Ruby for ```[1].max(2)``` and returns ```1```.
So, I think is ok to return one element.

src/array_util.rb
crisan-catalin: I used to have classes for better code management (in big projects).

alexandru-calinoiu: But this is not a big project, is always better to adapt to the problem.

Checkout how Matz (the creator of ruby) uses ruby to solve a simple problem https://www.youtube.com/watch?v=zEf6iUIkjf4

src/array_util.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def self.sum_even_numbers(numbers)
+    numbers.select(&:even?).sum
```
alexandru-calinoiu, line 4: You did not answer my question regarding nil guarding. 

A good approach is to init with a safe value if the received value is not great, you can do this with `||=` check out the docs and let's discuss this when we meet in person.

src/array_util.rb
alexandru-calinoiu: This is not what I meant, I was referring to:

```ruby
 +      firstNumber, secondNumber = secondNumber, firstNumber 
```

and

```ruby
secondNumber = firstNumber
firstNumber = x
```

src/array_util.rb
crisan-catalin: I know about ```||```, but I didn't know about```||=```, nice.

src/array_util.rb
crisan-catalin: Because if I have ```numbers = [1,2,3]``` the final result will be ```[3,1]``` instead of ```[3,2]``` without initial switch.
So, I check that the largest value is set to ```firstNumber```.

alexandru-calinoiu:
Closing this, it was settled.

PR #37, opened 2018-07-10
Alex

AlexCozma:
Done


MySolution.rb
```diff
@@ -0,0 +1,48 @@
+class MySolution
+
+    #problem1
+    def sumOfEven(array)
+        sum = 0
```
alexandru-calinoiu: So c#/java like :) can you come up with a 1 liner?

Steps:
- how can you drop `sum = 0` and `return sum`? do you even need a `return` statement in ruby?
- what are some alternatives to `for`?
- condense code.

MySolution.rb
```diff
@@ -0,0 +1,48 @@
...
+            if i.even?
+                sum += i
+            end
+        end
+        return sum
+    end
+    
+    #problem2
+    def getLargest2Numbers(array)
+        bigMax = -1
```
alexandru-calinoiu: The problem states `an array of integers` does not say `unsigned integers`. So this solution will fail for an array like `[-3, -2]`

MySolution.rb
```diff
@@ -0,0 +1,48 @@
+class MySolution
+
+    #problem1
+    def sumOfEven(array)
+        sum = 0
+        for i in array
+            if i.even?
```
alexandru-calinoiu: :+1: for using `even?`

MySolution.rb
```diff
@@ -0,0 +1,48 @@
...
+        end
+        return sum
+    end
+    
+    #problem2
+    def getLargest2Numbers(array)
+        bigMax = -1
+        smallMax = -1
+        for i in array
+            if i > bigMax and i > smallMax #if bigger then both
```
alexandru-calinoiu: This comments don't actually bring any benefit. Please write the code in such a way that it reads clearly and it does not require any explanation.

MySolution.rb
```diff
@@ -0,0 +1,48 @@
...
+        end
+        result = [smallMax, bigMax]
+        return result
+    end
+
+    #problem3
+    def hasDublicates(array)
+        #without temporary data structure to help, or sorting the problem
+        #goes to compare each element with each other
+        for i in 0..array.length * array.length
```
alexandru-calinoiu: Very cool solution.

But you technically cheated `Try to find a solution by executing just one loop.` you masked a loop in a loop with your 2 indexes.

alexandru-calinoiu (changes requested):
2 solutions are still pending and 1 can be improved.

levelboy:
@AlexCozma Before I even looked at the solution, I took a look over the tests. That doesn't look like TDD. 
There are three rules of TDD that drive the red green-refactor-cycle (taken from here: http://butunclebob.com/ArticleS.UncleBob.TheThreeRulesOfTdd)

1. You are not allowed to write any production code unless it is to make a failing unit test pass.
2. You are not allowed to write any more of a unit test than is sufficient to fail; and compilation failures are failures.
3. You are not allowed to write any more production code than is sufficient to pass the one failing unit test.

Please have a look at this video https://www.youtube.com/watch?v=qkblc5WRn-U 
See how the tests evolve from very simple to more complex, along as the code evolves also.

MySolution.rb
AlexCozma: Well I admit that C# is may favorite language :) and Ruby is truly new to me and I just got started learning it - And Yeah I can try to make it in one line.

- For dropping `sum = 0` and the `return sum` I can use an array function from ruby `array.sum` and to sum over only the even values something like `array.select{|v| v.even?}` (that looks like a LINQ from C#)
- If you referring to loop alternatives to `for` in ruby I know about `while`, `until`,`begin-while`

MySolution.rb
AlexCozma: Thanks :)

MySolution.rb
AlexCozma: You are right, I must had in mind the hypothesis from problem 3 after I read them. I'll add this Test Case and correct the solution

MySolution.rb
AlexCozma: Ok, I understood

MySolution.rb
AlexCozma: Well I admit that C# is may favorite language :) and Ruby is truly new to me and I just got started learning it - And Yeah I can try to make it in one line.

-For dropping `sum = 0` and the return sum I can use an array function from ruby `array.sum` and to sum over only the even values something like `array.select{|v| v.even?}` (that looks like a LINQ from C#)
-If you referring to loop alternatives to for in ruby I know about `while`, `until`,`begin-while`

MySolution.rb
AlexCozma: You're right, I've must had in mind the hypothesis from problem 3 after I read the problems. I'll correct it

MySolution.rb
AlexCozma: Thanks :)

MySolution.rb
AlexCozma: Actually yeah, the comment was redundant

MySolution.rb
AlexCozma: Thanks. 
Yes I'm aware that is technically a cheat, as I wrote in the comment I couldn't see anyway how to compare each element with each other. I'll try to find a different approach

.vscode/launch.json
```diff
@@ -0,0 +1,71 @@
+{
```
alexandru-calinoiu, line 1: I don't believe this file should have been commited.

MySolution.rb
```diff
@@ -0,0 +1,36 @@
...
+    def sumOfEven(array)
+        array.select{|v| v.even?}.sum
+    end
+    
+    #problem2
+    def getLargest2Numbers(array)
+        bigMax = -1/0.0
+        smallMax = -1/0.0
+        for i in array
+            if i > bigMax and i > smallMax
```
alexandru-calinoiu: Using `and` comes with a warning https://stackoverflow.com/questions/1426826/difference-between-and-and-in-ruby

MySolution.rb
```diff
@@ -0,0 +1,36 @@
...
+    
+    #problem2
+    def getLargest2Numbers(array)
+        bigMax = -1/0.0
+        smallMax = -1/0.0
+        for i in array
+            if i > bigMax and i > smallMax 
+                smallMax = bigMax
+                bigMax = i  
+            else
```
alexandru-calinoiu: Syntax can be shorten with `elsif`

MySolution.rb
```diff
@@ -0,0 +1,36 @@
...
+                if i > smallMax 
+                    smallMax = i 
+                end
+            end
+        end
+        result = [smallMax, bigMax]
+        return result
```
alexandru-calinoiu: Still using `return`'s :(

MySolution.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    #problem3
+    def hasDublicates(array)
+        for i in 0..array.length 
+            if array.select {|v| v == array[i]}.size > 1
```
alexandru-calinoiu: Still not right, `select` uses a loop in it's implementation therefor you still have a inner loop.

MySolution.rb
```diff
@@ -0,0 +1,36 @@
+class MySolution
+
+    #problem1
+    def sumOfEven(array)
+        array.select{|v| v.even?}.sum
```
alexandru-calinoiu, line 5: This looks like ruby now :+1: 

You can make it even nicer take a look at `inject`

MySolutionUnitTest.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+#problem 3
+    def test_hasDublicates_true
+        array = [5, 9, 1, 18, 6, 25, 16, 6]
+        assert_equal true,@unit.hasDublicates(array)
+    end
+
+    def test_hasDublicates_false
+        array = [5, 9, 1, 18, 6, 25, 16, 8]
```
alexandru-calinoiu: Your test case is not valid, the problem states `Given an array of numbers having size N in which every number is between 1 and N`

MySolution.rb
levelboy: Do you need the second condition here?

MySolution.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+    
+    #problem2
+    def getLargest2Numbers(array)
+        bigMax = -1/0.0
+        smallMax = -1/0.0
+        for i in array
```
levelboy: Ruby code should be more expressive. 
Confusion here is that usually, when you do a `for`, the `i` is the index. In this case, it's the actual item in the array.

MySolution.rb
levelboy: Do you need a variable named result here?
Do you need to explicitly specify `return` in ruby? (this was a previous question from @alexandru-calinoiu )

MySolution.rb
levelboy: This is still O(n^2)

MySolution.rb
```diff
@@ -0,0 +1,36 @@
...
+                end
+            end
+        end
+        result = [smallMax, bigMax]
+        return result
+    end
+
+    #problem3
+    def hasDublicates(array)
+        for i in 0..array.length
```
levelboy: Emphasizing the confusion again for how you use `i`. This time, it's an item in an array of indices, used to access the array in the inner loop.

MySolution.rb
```diff
@@ -0,0 +1,36 @@
...
+        end
+        result = [smallMax, bigMax]
+        return result
+    end
+
+    #problem3
+    def hasDublicates(array)
+        for i in 0..array.length 
+            if array.select {|v| v == array[i]}.size > 1
+                return true
```
levelboy: See how `return` works in a ruby function

MySolutionUnitTest.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    def test_getLargest2Numbers_negativeValues
+        array = [-3, -2]
+        assert_equal array,@unit.getLargest2Numbers(array)
+    end
+
+#problem 3
+    def test_hasDublicates_true
+        array = [5, 9, 1, 18, 6, 25, 16, 6]
```
levelboy: The input to the 3rd problem does not respect the problem statement:
`Given an array of numbers having size N in which every number is between 1 and N`

.vscode/launch.json
levelboy, line 1: What's up with this file? :)

levelboy (changes requested):
Please see requested changes.

alexandru-calinoiu (changes requested):

MySolution.rb
AlexCozma: Giving a second thought, yes it's actually redundant.

MySolution.rb
AlexCozma: Yes, it is, this if from my habit that first choice for a variable in for to be named `i`. In this case it is confusing, I'll give it a suggestive name

MySolution.rb
AlexCozma: In ruby, no I don't need a explicit result. I didn't know how to return two variables at once, that's why I used a variable. But yes I can leave it as `[smallMax, bigMax]`

MySolutionUnitTest.rb
AlexCozma: You're right I skip notice of it, I'll correct it

.vscode/launch.json
AlexCozma: It's the debugger configuration file from visual studio code, is generated in the workspace folder where also is my local repository. Sorry for that

AlexCozma: It's the debugger configuration file from visual studio code, is generated in the workspace folder where also is my local repository. Sorry for that

MySolution.rb
AlexCozma: I didn't know that. Still the second condition is redundant

MySolution.rb
AlexCozma: OK I'll change it

MySolution.rb
AlexCozma: I'll remove it

MySolutionUnitTest.rb
AlexCozma: I'll correct it.

MySolutionUnitTest.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    def test_getLargest2Numbers_BigMaxFirst
+        array = [5, 9, 1, 19, 14, 3, 10, 25, 9]
+        result = [19,25]
+        assert_equal result,@unit.getLargest2Numbers(array)
+    end
+
+    def test_getLargest2Numbers_negativeValues
+        array = [-3, -2]
```
levelboy, line 41: Does this test work if you put set the array to: [-2, -3]?

AlexCozma: Yes it passes

PR #38, opened 2018-07-30
Tasks completed

BogdanH95:


src/Task1KtTest.kt
```diff
@@ -0,0 +1,16 @@
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.Test
+import test.agileFreaks.Task1
+
+internal class Task1KtTest {
+
+    @Test
+    fun onEmptyArray() {
```
alexandru-calinoiu: You wrote to many tests for the first iteration. You should have only 1 failing test at once.

src/Tasks.kt
```diff
@@ -0,0 +1,10 @@
+package test.agileFreaks
+
+fun IntArray.sumOfEven():Int {
```
alexandru-calinoiu: Why refactor? What was wrong with the first implementation?

src/Task1Test.kt
```diff
@@ -0,0 +1,16 @@
...
+
+internal class Task1Test {
+
+    @Test
+    fun onEmptyArray() {
+        assertEquals(0, intArrayOf().sumOfEven())
+    }
+
+    @Test
+    fun onArrayOfInt(){
```
alexandru-calinoiu: Missing some test cases, for example what happens when all numbers are odd?

src/Tasks.kt
```diff
@@ -0,0 +1,10 @@
...
+    var sum =0
+    this.forEach {
+        if (it % 2 == 0) sum += it
```
alexandru-calinoiu: If you really want to refactor, how about an extension method to encapsulate the even logic?

src/Task2Test.kt
```diff
@@ -0,0 +1,37 @@
...
+import kotlin.test.*
+
+
+internal class Task2Test {
+
+    @Test
+    //Exceptional case where index -1 is returned
+    // to show that an array of length of less than
+    // 2 was given as input.
+    fun onEmptyArray() {
```
alexandru-calinoiu: Same problem as for problem 1.

The whole point of the TDD is the play between tests and code, you write a little bit of code and then some code to make the test pass and then refactor if the case might be and then start over.

src/Task2Test.kt
```diff
@@ -0,0 +1,37 @@
...
+        assertEquals(-1, max1)
+        assertEquals(-1, max2)
+    }
+
+    @Test
+    //Usual scenario.
+    fun onArrayOfInt(){
+        val (max1,max2) = intArrayOf(1,2,10,3,4,11,5,6,7,8,9).firstTwoLargestIndexes()
+
+        assertEquals(5, max1)
```
alexandru-calinoiu: The problem statements asks for the first and second max, not for their indexes.

src/Tasks.kt
```diff
@@ -9,5 +9,16 @@ fun IntArray.sumOfEven():Int {
...
 
 fun IntArray.firstTwoLargestIndexes():IntArray {
-    return intArrayOf(-2,-2)
+    if(this.count() < 2 ) return intArrayOf(-1,-1)
+
+    var max = Int.MIN_VALUE
+    var theOtherOne = Int.MIN_VALUE
+
+    this.forEach {
+        if(it > max){
```
alexandru-calinoiu: What about this test array:

[10, 5, 1]

Does your solution work?

src/Task3Test.kt
```diff
@@ -7,17 +7,17 @@ internal class Task3Test {
...
     @Test
     // Same bad input scenario.
-    fun onSingleElementArray() = assertTrue(!intArrayOf().hasDuplicates())
+    fun onSingleElementArray() = assertTrue(!hasDuplicates(intArrayOf(1)))
 
     @Test
-    //Usual scenario.
-    fun onNoDuplicatesArrayOfInt() = assertTrue(intArrayOf(1,2,3,4,5,6,7,8,9,10).hasDuplicates())
+    //Usual scenario without duplicates.
+    fun onNoDuplicatesArrayOfInt() = assertTrue(hasDuplicates(intArrayOf(1,2,3,4,5,6,7,8,9,10)))
```
alexandru-calinoiu: Well actually this array has no duplicates.

Your solution seems promising but is wrong.

alexandru-calinoiu (changes requested):
Only the first problem was correctly resolved, please review the comments.

src/Tasks.kt
BogdanH95: The reason I've refactored it is the complexity and the unnecessary reiteration that was done after filtering the elements of the array, that one was executing two for loops, one to filter all the elements and another one to sum the remaining elements.

src/Task2Test.kt
BogdanH95: The problem as it is now states that I must identify the largest 2 numbers. I've identified their identities in the array instead of returning them. I'll change the implementation to return the numbers.

src/Tasks.kt
BogdanH95: Awesome idea, thanks for the info!

src/Tasks.kt
BogdanH95: No, it doesn't

src/Task3Test.kt
BogdanH95: A bug in code backed up by a bug in the test.. 👍 Fixed

src/Task1.kt
```diff
@@ -0,0 +1,12 @@
+
+fun IntArray.sumOfEven():Int {
+    var sum = 0
+    this.forEach{
+    if (it.isEven()) sum += it
```
pirvudoru: This is an article about coding style and why it matters. https://www.smashingmagazine.com/2012/10/why-coding-style-matters/

It references web but you can check out https://android.github.io/kotlin-guides/style.html#indentation

src/Task1.kt
```diff
@@ -0,0 +1,12 @@
+
+fun IntArray.sumOfEven():Int {
+    var sum = 0
+    this.forEach{
```
pirvudoru: We can do better. Can you write this as a 1 liner ?

src/Task2.kt
```diff
@@ -0,0 +1,13 @@
+
+fun IntArray.firstTwoLargestIndexes():IntArray {
+    var max = Int.MIN_VALUE
+    var theOtherOne = Int.MIN_VALUE
+
+    this.forEach {
+        if(it > max){
+            theOtherOne = max
+            max =  it
+        }else if (it > theOtherOne ) theOtherOne  = it
```
pirvudoru: Re: coding style

src/Task2.kt
```diff
@@ -0,0 +1,13 @@
+
+fun IntArray.firstTwoLargestIndexes():IntArray {
```
pirvudoru: Function name does not reflect result.

src/Task2Test.kt
```diff
@@ -0,0 +1,42 @@
...
+    // 2 was given as input.
+    fun onEmptyArray() {
+        val (max1,max2) = intArrayOf().firstTwoLargestIndexes()
+
+        assertEquals(Int.MIN_VALUE, max1)
+        assertEquals(Int.MIN_VALUE, max2)
+    }
+
+    @Test
+    // Same bad input scenario.
```
pirvudoru: Not helpful. Scenario (AAA) should be reflected by test fun name. 

When to use comments: https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/

src/Task3.kt
```diff
@@ -0,0 +1,16 @@
+
+//Made it function in order to manipulate the array
+fun hasDuplicates(array : IntArray): Boolean {
+    if(array.count() < 2) return false
+
+    array.forEach{
+        val indexNormalized = Math.abs(it) -1
```
pirvudoru: Why do you need to abs it ? Problem states:
`Given an array of numbers having size N in which every number is between 1 and N`

src/Task3Test.kt
```diff
@@ -0,0 +1,23 @@
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.Test
+
+internal class Task3Test {
+
+    @Test
+    //Exceptional case where index -1 is returned
+    // to show that an array of length of less than
+    // 2 was given as input.
+    fun onEmptyArray() = assertTrue(!hasDuplicates(intArrayOf()))
```
pirvudoru: Why not `assertFalse(hasDuplicates(intArrayOf())` ?

src/Task3Test.kt
```diff
@@ -0,0 +1,23 @@
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.Test
+
+internal class Task3Test {
+
+    @Test
+    //Exceptional case where index -1 is returned
+    // to show that an array of length of less than
+    // 2 was given as input.
```
pirvudoru: What does this comment even mean ? it is the same for Task2 test1. This method returns bool, never -1?

src/Task3.kt
```diff
@@ -0,0 +1,16 @@
+
+//Made it function in order to manipulate the array
+fun hasDuplicates(array : IntArray): Boolean {
+    if(array.count() < 2) return false
```
pirvudoru: Do you really need this check ?

pirvudoru (changes requested):

pirvudoru:
Please be consistent with commit messages. https://chris.beams.io/posts/git-commit/

src/Task1Test.kt
```diff
@@ -0,0 +1,35 @@
...
+        assertEquals(6, intArrayOf(1,1,2,1,2,1,1,2,1,1).sumOfEven())
+    }
+
+    @Test
+    fun sumOfEvenOnArrayOfEvenInt(){
+        assertEquals(6, intArrayOf(2,0,0,-2,0,4,2,6,-6).sumOfEven())
+    }
+
+    @Test
+    fun sumOfEvenOnArrayOfOddInt(){
```
alexandru-calinoiu, line 32: For future reference you can write more meningfull names for your tests with ``

For example:

```kotlin
fun `sum will ignore odd numbers`() { /* code */ }
```

src/Task3.kt
BogdanH95: The implementation I've used makes them negative to check them as seen

src/Task3Test.kt
alexandru-calinoiu: Proves the point that comment should rarely be used as they fall out of sync with the code.

Ideally the code should be so clear that it should not require a comment.

src/Task3Test.kt
```diff
@@ -0,0 +1,23 @@
...
+    fun onEmptyArray() = assertTrue(!hasDuplicates(intArrayOf()))
+
+    @Test
+    // Same bad input scenario.
```
alexandru-calinoiu: More like: 'same bad comment', the input is fine you should be able to pass and array with 1 element and it will obviously have no duplicates :)

src/Task3.kt
```diff
@@ -0,0 +1,16 @@
+
+//Made it function in order to manipulate the array
```
alexandru-calinoiu: This comments man, they are so confusing :)

src/Task2Test.kt
```diff
@@ -0,0 +1,42 @@
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.Test
+
+internal class Task2Test {
+
+    @Test
+    //Exceptional case where index -1 is returned
```
alexandru-calinoiu: Strange comment. You keep refering at this as indexes but they are actually values.

alexandru-calinoiu (changes requested):
The problems seem to be solved this time but the work is very sloppy a lot of strange comments, badly formatted code and not relevant function naming.

Please review solutions, use you IDE to format code, and try to express your self with no comments.

src/Task1.kt
BogdanH95: You mean like the [first implementation ](https://github.com/Agilefreaks/Test/pull/38/commits/0696d02cb9368c31cfcd1230510a3f39ad3f44cd) was? I thought that it's better, but I'm just skipping a for statement , I'll change with that.

src/Task1.kt
```diff
@@ -0,0 +1,6 @@
+
+fun IntArray.sumOfEven ():Int = this.filter {it.isEven()}.sum()
```
alexandru-calinoiu, line 2: This looks very nice,

But the style man, it hurts my eyes when I read it.

src/Task1Test.kt
```diff
@@ -0,0 +1,35 @@
...
+        assertTrue(2.isEven())
+    }
+
+    @Test
+    fun isEvenOnOddNumber(){
+        assertFalse(1.isEven())
+    }
+
+    @Test
+    fun sumOfEvenOnEmptyArray() {
```
alexandru-calinoiu, line 17: Space

src/Task1Test.kt
```diff
@@ -0,0 +1,35 @@
...
+        assertEquals(0, intArrayOf().sumOfEven())
+    }
+
+    @Test
+    fun sumOfEvenOnArrayOfInt(){
```
alexandru-calinoiu, line 22: No space

alexandru-calinoiu (approved):
This looks good, congratulations.

alexandru-calinoiu:
O sa inchid pr'ul urmeaza sa followup via email.

PR #40, opened 2018-09-26
Solution by robert

robertlisaru:


alexandru-calinoiu:
Please remove .idea from pr, I use Code to edit js so your .idea settings are of no value.

See ignore file https://github.com/github/gitignore/blob/9d7ff09c7d38dce9ef03e7ea4dc908a622546757/Global/JetBrains.gitignore

solution/test/test-has-duplicates.js
```diff
@@ -0,0 +1,19 @@
...
+    const numbers = [6, 2, 5, 4, 3, 6, 1];
+    assert(hasDuplicates(numbers) == true);
+});
+
+it("test duplicates in single number array", () => {
+    const numbers = [1];
+    assert(hasDuplicates(numbers) == false);
+});
+
+//I'll skip the input validation tests for now
```
pirvudoru: No need for this comment. Useful comments describe "WHY" not "WHAT"

solution/test/test-has-duplicates.js
```diff
@@ -0,0 +1,19 @@
+const hasDuplicates = require('../src/has-duplicates');
+const assert = require('assert');
+
+it("test when no duplicates", () => {
+    const numbers = [3, 2, 1, 4, 7, 6, 5];
+    assert(hasDuplicates(numbers) == false);
+});
+
+it("test when duplicates", () => {
```
pirvudoru, line 9: How will the following test case affect your solution?  ?
```
[1,1,4,4]
```

solution/src/sum-evens.js
```diff
@@ -0,0 +1,8 @@
+function sumEvens(numbers) {
+    if (!Array.isArray(numbers)) {
+        return undefined;
+    }
```
pirvudoru: If you did not guard in `hasDuplicates` why guard here?

solution/test/test-has-duplicates.js
```diff
@@ -0,0 +1,19 @@
...
+    const numbers = [6, 2, 5, 4, 3, 6, 1];
+    assert(hasDuplicates(numbers) == true);
+});
+
+it("test duplicates in single number array", () => {
+    const numbers = [1];
+    assert(hasDuplicates(numbers) == false);
```
pirvudoru: Why `assert(functionCall() == false)` and not just `assert(functionCall())` ?

solution/coverage/largest-2.js.html
```diff
@@ -0,0 +1,102 @@
...
+<span class="cline-any cline-yes">5x</span>
+<span class="cline-any cline-neutral">&nbsp;</span>
+<span class="cline-any cline-neutral">&nbsp;</span>
+<span class="cline-any cline-yes">1x</span>
+<span class="cline-any cline-neutral">&nbsp;</span>
+<span class="cline-any cline-neutral">&nbsp;</span>
+<span class="cline-any cline-yes">1x</span></td><td class="text"><pre class="prettyprint lang-js">function largest2(numbers) {
+    largest = [Number.MIN_VALUE, Number.MIN_VALUE];
+    for (i = 0; i &lt; numbers.length; i++) {
+        if (numbers[i] &gt; largest[0]) {
```
pirvudoru: Too much noise with files that don't need to be here. Please clean-up so that your solution so it's more clear.

solution/test/test-sum-evens.js
```diff
@@ -0,0 +1,26 @@
...
+});
+
+it("sum evens when non array", () => {
+    assert.equal(sumEvens("not an array"), undefined);
+});
+
+it("sum evens when non integer values", () => {
+    const nonIntegers = [10, "definitely not an integer", 2, 3, 3.3];
+    assert.equal(sumEvens(nonIntegers), undefined);
+});
```
pirvudoru: image

solution/test/test-largest-2.js
```diff
@@ -0,0 +1,11 @@
+const largest2 = require('../src/largest-2');
+const assert = require('assert');
+
+it("largest 2", () => {
+    const numbers = [1, 2, 3, 100, 4, 5, 6, 200];
+    let largest = largest2(numbers);
```
pirvudoru, line 6: How will the following test case affect your solution? ?

```
[1, 2, 5, 4, 3]
```

solution/package.json
```diff
@@ -0,0 +1,16 @@
...
+    "test": "mocha",
+    "test-with-coverage": "nyc --reporter=html --reporter=text mocha"
+  },
+  "author": "Robert",
+  "license": "ISC",
+  "devDependencies": {
+    "mocha": "^5.2.0",
+    "nyc": "^13.0.1"
+  }
+}
```
pirvudoru, line 16: How did you generate this file?

solution/src/sum-evens.js
```diff
@@ -0,0 +1,8 @@
...
+    return numbers.filter((x) => x % 2 == 0).reduce((x, y) => x + y, 0);
```
pirvudoru: :100:

pirvudoru (changes requested):
Nice first attempt but it requires more work. please see comments

solution/test/test-sum-evens.js
pirvudoru: also the problem states:

Given an **integer array** of size N, compute the sum of all even numbers in this array

pirvudoru:
Suggested reading: https://chris.beams.io/posts/git-commit/

solution/package.json
robertlisaru: I run `npm init`, it asked me for most details.
Then I installed and saved _mocha_ test runner, `npm install --save-dev mocha`,
and _instanbul_ for test coverage reports `npm install --save-dev nyc` , which were added to "devDependencies".
Then I added a new script "test-with-coverage" which runs nyc and generates coverage report.
Is there a problem with the file?

solution/test/test-has-duplicates.js
robertlisaru: I understand. Will clean-up. Thanks!

solution/test/test-has-duplicates.js
robertlisaru: Sorry, but is that correct? Shouldn't it be `assert( ! functionCall())` in order to be the same as `assert(functionCall() == false)` ?

solution/src/sum-evens.js
robertlisaru: Just for the sake of exercising. But I'll assume input is valid (as stated in the problem description) and remove all the guards for a cleaner solution. Thank you!

solution/coverage/largest-2.js.html
robertlisaru: In the mail I got it says 

> Te rog sa incluzi un raport de test coverage in pr ca sa demonstrezi ca ai 100% coverage.

I wasn't sure what format should the report follow, so I included the _HTML_ report generated by _nyc_. 
Could you provide a few more details about how the report should look like?

solution/test/test-sum-evens.js
robertlisaru: I understand. Will remove unnecessary tests.

solution/test/test-largest-2.js
robertlisaru: Turns out my solution is wrong. Will fix.

solution/test/test-has-duplicates.js
robertlisaru: Ah, my solution is naive, should have thought more thoroughly. Will fix.

robertlisaru (commented):
This is the bit-vector approach, but I didn't explicitly allocate a bit-vector. The bit-vector is encoded in the input array by making the integers negative. So maybe I didn't break the rules?

solution/src/has-duplicates.js
```diff
@@ -0,0 +1,11 @@
+function hasDuplicates(numbers) {
+    for (let i = 0; i < numbers.length; i++) {
+        if (numbers[Math.abs(numbers[i])] < 0) {
+            return true;
+        }
+        numbers[Math.abs(numbers[i])] *= -1;
```
alexandru-calinoiu: This is a good solution.

You can make it easier to read by removing some duplication, no point on getting the absolute value two times.

solution/src/largest-2.js
```diff
@@ -0,0 +1,15 @@
+function largest2(numbers) {
+    let max = Number.MIN_VALUE;
+    let max2 = Number.MIN_VALUE;
+    for (let i = 0; i < numbers.length; i++) {
+        if (numbers[i] > max) {
```
alexandru-calinoiu, line 5: Now that you have decent test coverage you can easily refactor this method, one thing you could to is remove duplication by removing the `number[i]` duplication. You've used `.filter` in the first problem can you also improve this problem?

solution/src/sum-evens.js
alexandru-calinoiu: I will go with 90 not 100, it will be :100: if the variable names won't be some lame, `x` and `y`? Can you please be more expressive.

solution/test/test-has-duplicates.js
```diff
@@ -0,0 +1,22 @@
...
+});
+
+it("test when duplicates", () => {
+    const numbers = [6, 2, 5, 4, 3, 6, 1];
+    assert(hasDuplicates(numbers));
+});
+
+it("test duplicates in single number array", () => {
+    const numbers = [1];
+    assert(!hasDuplicates(numbers));
```
alexandru-calinoiu, line 16: Negations are hard to reason about, can you write an assertion that will not require you to negate the result.

solution/test/test-largest-2.js
```diff
@@ -0,0 +1,16 @@
+const largest2 = require('../src/largest-2');
```
alexandru-calinoiu, line 1: you should have probably have a test with negative number

solution/src/has-duplicates.js
robertlisaru: That's right. Will fix.

solution/src/largest-2.js
robertlisaru: I already have commited a solution using `.filter`
See commit: https://github.com/Agilefreaks/Test/pull/40/commits/0d0c716024591acea0d40ee52f827021b416ed44
It looks like this:
```javascript
function largest2(numbers) {
    let max = Math.max(...numbers);
    let max2 = Math.max(...numbers.filter((x) => x != max));
    return [max, max2];
}
```

solution/test/test-has-duplicates.js
robertlisaru: `assert(hasDuplicates(numbers) == false)` could be an alternative, but Doru says:
> Why `assert(functionCall() == false)` and not just `assert(functionCall())` ?

Another alternative:
`assert.equal(hasDuplicates(numbers), false)`

If something like "assertFalse" is what you're thinking of, I cannot find such a function:
https://nodejs.org/api/assert.html

solution/src/sum-evens.js
robertlisaru: I've renamed to:
```javascript
return numbers.filter((number) => number % 2 == 0)
        .reduce((previousSum, currentValue) => previousSum + currentValue, 0);
```
Perhaps a bit better. Let me know if it's still not suggestive. Thank you!

solution/test/test-largest-2.js
robertlisaru: Right. Will include one. Thanks!

alexandru-calinoiu:
Will close this and followup via email.

PR #41, opened 2018-09-30
Solution proposal

alex-damian-negru:
This is the solution for the requested exercises built using Ruby and TDD methodology. The testing framework used was the behavior-driven RSpec.

Looking forward to hearing from you.


problem_three.rb
```diff
@@ -0,0 +1,19 @@
+# Given an array of numbers having size N in which every number is between 1 and N, determine if there are any duplicates in it. 
+# - Do not sort the array or use bit vectors.
+# - Do not use 2 nesting for loops since the time complexity would be high.
+# - You are allowed to destroy the array if you like.
+# - Try to work within the array without using any other temporary data structures.
+# - Try to find a solution by executing just one loop.
+
+class ProblemThree
+  def duplicates?(array)
+    for i in 1..array.length do
```
drqCode, line 10: Could find an alternative for iterating with `for`? This solution isn't specific to Ruby world...

problem_two.rb
```diff
@@ -0,0 +1,10 @@
+# Given an integer array of size N, identify the largest 2 numbers in this array. Do not sort the array.
+
+class ProblemTwo
+  def max_two(array)
+    raise 'Invalid input: array must contain at least two numbers' \
+      if array.length < 2
```
drqCode, line 6: There's no assumption in the problem statement that the array MUST contain at least two numbers.

alexandru-calinoiu:
@alex-damian-negru I can't check the commit messages as they seem to originate from  a different branch, without the commits is impossible to see that proper tdd was done.

screenshot from 2018-10-01 10-06-54

spec/problem_one_spec.rb
```diff
@@ -0,0 +1,33 @@
+require_relative '../problem_one'
+
+describe ProblemOne do
+  it 'should compute for a set of even numbers' do
+    p_one = ProblemOne.new
```
drqCode, line 5: Could you use `let`s? Check this: https://relishapp.com/rspec/rspec-core/v/2-5/docs/helper-methods/let-and-let

problem_two.rb
```diff
@@ -0,0 +1,10 @@
...
+
+    array.max(2)
```
drqCode, line 8: This is debatable. The statement doesn't say to identify the largest 2 elements of the array. I think there's a difference between `numbers` and `elements`, don't you think?

alex-damian-negru:
To @alexandru-calinoiu, @drqCode:

I had to create a new pull request in order to fix the commit history: https://github.com/Agilefreaks/Test/pull/42

I will close this pull request as it is no longer relevant.

PR #42, opened 2018-10-01
Solution proposal (commit history fixed)

alex-damian-negru:
In order to fix the commit history being hidden, I had to create a new pull request. I apologize for any inconveniences. 

I have also implemented the requested changes from the comments in the previous pull request (https://github.com/Agilefreaks/Test/pull/41).

### DIFF
--- a/problem_one.rb
+++ b/problem_one.rb
@@ -2,10 +2,6 @@
 
 class ProblemOne
   def sum_even(array)
-    sum = 0
-    array.each do |item|
-      sum += item if item.even?
-    end
-    sum
+    array.select(&:even?).reduce(:+)
   end
 end
--- a/problem_one.rb
+++ b/problem_one.rb
@@ -2,6 +2,6 @@
 
 class ProblemOne
   def sum_even(array)
-    array.select(&:even?).reduce(:+)
+    array.select(&:even?).reduce(:+) || 0
   end
 end
--- a/problem_one.rb
+++ b/problem_one.rb
@@ -2,6 +2,10 @@
 
 class ProblemOne
   def sumEven(array)
-    return 0
+    sum = 0
+    array.each do |item|
+      sum += item if item % 2 == 0
+    end
+    return sum
   end
 end
--- a/problem_one.rb
+++ b/problem_one.rb
@@ -1,0 +1,1 @@
+# Given an integer array of size N, compute the sum of all even numbers in this array.
--- a/problem_one.rb
+++ b/problem_one.rb
@@ -1,11 +1,11 @@
 # Given an integer array of size N, compute the sum of all even numbers in this array.
 
 class ProblemOne
-  def sumEven(array)
+  def sum_even(array)
     sum = 0
     array.each do |item|
-      sum += item if item % 2 == 0
+      sum += item if item.even?
     end
-    return sum
+    sum
   end
 end
--- a/problem_one.rb
+++ b/problem_one.rb
@@ -1,1 +1,7 @@
 # Given an integer array of size N, compute the sum of all even numbers in this array.
+
+class ProblemOne
+  def sumEven(array)
+    return 0
+  end
+end
--- a/problem_one.rb
+++ b/problem_one.rb
@@ -1,7 +1,5 @@
 # Given an integer array of size N, compute the sum of all even numbers in this array.
 
-class ProblemOne
-  def sum_even(array)
-    array.select(&:even?).reduce(0) { |sum, num| sum + num }
-  end
+def sum_even(array)
+  array.select(&:even?).reduce(0) { |sum, num| sum + num }
 end
--- a/problem_one.rb
+++ b/problem_one.rb
@@ -2,6 +2,7 @@
 
 class ProblemOne
   def sum_even(array)
-    array.select(&:even?).reduce(:+) || 0
+    sum = array.select(&:even?).reduce(:+)
+    sum.nil? ? 0 : sum
   end
 end
--- a/problem_one.rb
+++ b/problem_one.rb
@@ -2,7 +2,6 @@
 
 class ProblemOne
   def sum_even(array)
-    sum = array.select(&:even?).reduce(:+)
-    sum.nil? ? 0 : sum
+    array.select(&:even?).reduce(0) { |sum, num| sum + num }
   end
 end
--- a/problem_three.rb
+++ b/problem_three.rb
@@ -7,12 +7,12 @@
 
 class ProblemThree
   def duplicates?(array)
-    for i in 1..array.length do
-      raise "array#{[i - 1]}=#{array[i - 1]} outside 1 and #{array.length}" \
-        if array[i - 1].abs > array.length
-      return true if array[array[i - 1].abs - 1] < 0
+    array.each_with_index do |_, index|
+      raise "array#{[index - 1]}=#{array[index - 1]} outside 1 and #{array.length}" \
+        if array[index - 1].abs > array.length
+      return true if array[array[index - 1].abs - 1] < 0
 
-      array[array[i - 1].abs - 1] > 0 && array[array[i - 1].abs - 1] *= -1
+      array[array[index - 1].abs - 1] > 0 && array[array[index - 1].abs - 1] *= -1
     end
     false
   end
--- a/problem_three.rb
+++ b/problem_three.rb
@@ -1,0 +1,6 @@
+# Given an array of numbers having size N in which every number is between 1 and N, determine if there are any duplicates in it. 
+# - Do not sort the array or use bit vectors. 
+# - Do not use 2 nesting for loops since the time complexity would be high. 
+# - You are allowed to destroy the array if you like. 
+# - Try to work within the array without using any other temporary data structures. 
+# - Try to find a solution by executing just one loop.
--- a/problem_three.rb
+++ b/problem_three.rb
@@ -7,11 +7,14 @@
 
 class ProblemThree
   def duplicates?(array)
+    has_duplicate = false
     array.each_with_index do |_, index|
-      return true if array[array[index - 1].abs - 1] < 0
-
+      if array[array[index - 1].abs - 1] < 0
+        has_duplicate = true
+        break
+      end
       array[array[index - 1].abs - 1] > 0 && array[array[index - 1].abs - 1] *= -1
     end
-    false
+    has_duplicate
   end
 end
--- a/problem_three.rb
+++ b/problem_three.rb
@@ -6,14 +6,14 @@
 # - Try to find a solution by executing just one loop.
 
 class ProblemThree
-  def find_duplicates(array)
+  def duplicates?(array)
     for i in 1..array.length do
-      if array[array[i-1].abs-1] > 0
-        array[array[i-1].abs-1] = -array[array[i-1].abs-1]
-      else
-        return 'true'
+      return true if array[array[i - 1].abs - 1] < 0
+
+      if array[array[i - 1].abs - 1] > 0
+        array[array[i - 1].abs - 1] *= -1
       end
     end
-    return 'false'
+    false
   end
 end
--- a/problem_three.rb
+++ b/problem_three.rb
@@ -9,7 +9,7 @@ class ProblemThree
   def duplicates?(array)
     has_duplicate = false
     array.each_with_index do |_, index|
-      if array[array[index - 1].abs - 1] < 0
+      if array[array[index - 1].abs - 1].negative?
         has_duplicate = true
         break
       else
--- a/problem_three.rb
+++ b/problem_three.rb
@@ -9,11 +9,12 @@ class ProblemThree
   def duplicates?(array)
     has_duplicate = false
     array.each_with_index do |_, index|
-      if array[array[index - 1].abs - 1].negative?
+      value = array[index - 1].abs - 1
+      if array[value].negative?
         has_duplicate = true
         break
       else
-        array[array[index - 1].abs - 1] *= -1
+        array[value] *= -1
       end
     end
     has_duplicate
--- a/problem_three.rb
+++ b/problem_three.rb
@@ -8,8 +8,6 @@
 class ProblemThree
   def duplicates?(array)
     array.each_with_index do |_, index|
-      raise "array#{[index - 1]}=#{array[index - 1]} outside 1 and #{array.length}" \
-        if array[index - 1].abs > array.length
       return true if array[array[index - 1].abs - 1] < 0
 
       array[array[index - 1].abs - 1] > 0 && array[array[index - 1].abs - 1] *= -1
--- a/problem_three.rb
+++ b/problem_three.rb
@@ -12,8 +12,9 @@ class ProblemThree
       if array[array[index - 1].abs - 1] < 0
         has_duplicate = true
         break
+      else
+        array[array[index - 1].abs - 1] *= -1
       end
-      array[array[index - 1].abs - 1] > 0 && array[array[index - 1].abs - 1] *= -1
     end
     has_duplicate
   end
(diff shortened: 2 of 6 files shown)


problem_three.rb
```diff
@@ -0,0 +1,19 @@
...
+# - Do not sort the array or use bit vectors.
+# - Do not use 2 nesting for loops since the time complexity would be high.
+# - You are allowed to destroy the array if you like.
+# - Try to work within the array without using any other temporary data structures.
+# - Try to find a solution by executing just one loop.
+
+class ProblemThree
+  def duplicates?(array)
+    array.each_with_index do |_, index|
+      raise "array#{[index - 1]}=#{array[index - 1]} outside 1 and #{array.length}" \
```
pirvudoru: Why do you need this guard in place?

spec/problem_one_spec.rb
```diff
@@ -0,0 +1,30 @@
+require_relative '../problem_one'
+
+describe ProblemOne do
+  let(:p_one) { $p_one = ProblemOne.new }
+
+  it 'should compute for a set of even numbers' do
```
pirvudoru: Further reading: http://www.betterspecs.org/#should

spec/problem_two_spec.rb
```diff
@@ -0,0 +1,15 @@
+require_relative '../problem_two'
+
+describe ProblemTwo do
+  let(:p_two) { $p_two = ProblemTwo.new }
```
pirvudoru: Why do you need to do `$p_two = ...` ?

spec/problem_one_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  it 'should compute for a set with negative numbers' do
+    test_input = [7, -2, 3, 16, -10, -21, 100, 57, 131]
+    expect(p_one.sum_even(test_input)).to eq(104)
+  end
+
+  it 'should compute for an empty sum' do
+    test_input = []
+    expect(p_one.sum_even(test_input)).to eq(0)
```
pirvudoru: Further reading: https://www.justinweiss.com/articles/writing-better-tests-with-the-three-phase-pattern/

problem_three.rb
```diff
@@ -0,0 +1,19 @@
+# Given an array of numbers having size N in which every number is between 1 and N, determine if there are any duplicates in it. 
+# - Do not sort the array or use bit vectors.
+# - Do not use 2 nesting for loops since the time complexity would be high.
+# - You are allowed to destroy the array if you like.
+# - Try to work within the array without using any other temporary data structures.
+# - Try to find a solution by executing just one loop.
+
+class ProblemThree
+  def duplicates?(array)
```
pirvudoru: :+1:

spec/problem_three_spec.rb
```diff
@@ -0,0 +1,20 @@
+require_relative '../problem_three'
+
+describe ProblemThree do
+  let(:p_three) { $p_three = ProblemThree.new }
+  
+  it 'should determine if there are duplicates' do
+    test_input = [3, 4, 1, 6, 4, 1, 7, 9, 3, 10]
+    expect(p_three.duplicates?(test_input)).to eq(true)
```
pirvudoru: Further reading: http://www.betterspecs.org/#subject

problem_three.rb
```diff
@@ -0,0 +1,19 @@
...
+    array.each_with_index do |_, index|
+      raise "array#{[index - 1]}=#{array[index - 1]} outside 1 and #{array.length}" \
+        if array[index - 1].abs > array.length
+      return true if array[array[index - 1].abs - 1] < 0
```
pirvudoru: It's always better to keep a single return per function and avoid as much as possible short-circuiting the function with the "return" keyword.

problem_three.rb
```diff
@@ -0,0 +1,19 @@
...
+
+      array[array[index - 1].abs - 1] > 0 && array[array[index - 1].abs - 1] *= -1
```
pirvudoru: This is not better refactor. It is less code, but remember we are basically writing software mainly for humans, not machines. Someone else will have to maintain the code you write and you don't want to make his/her life hard (say you should be maintaining someone else's code)

See https://i2.wp.com/commadot.com/wp-content/uploads/2009/02/wtf.png?w=550

problem_one.rb
```diff
@@ -0,0 +1,7 @@
+# Given an integer array of size N, compute the sum of all even numbers in this array.
+
+class ProblemOne
+  def sum_even(array)
+    array.select(&:even?).reduce(:+) || 0
```
pirvudoru: How did you came about to using "reduce" ? 
Can we avoid `|| 0` ?

pirvudoru (changes requested):
Further reading: https://chris.beams.io/posts/git-commit/

problem_three.rb
alex-damian-negru: If the current value in the loop is out of bounds, the algorithm will fail, therefore it makes sense to stop further looping and point the developer to the error.

spec/problem_two_spec.rb
alex-damian-negru: It was requested by @drqCode in the previous pull request: 
> Could you use `let`s? Check this: https://relishapp.com/rspec/rspec-core/v/2-5/docs/helper-methods/let-and-let

problem_three.rb
pirvudoru: Given an array of numbers having size N in which **every number is between 1 and N**, determine if there are any duplicates in it. You are allowed to destroy the array if you like. Do not sort the array or use bit vectors. Try to work within the array without using any other temporary data structures. Do not use 2 nesting for loops since the time complexity would be high. Try to find a solution by executing just one loop.

alex-damian-negru: Removed redundant guard: https://github.com/Agilefreaks/Test/pull/42/commits/9b512f898bb4faa99d019d4f7d966489d0cdb9c4

spec/problem_one_spec.rb
alex-damian-negru: Changed descriptions to match the stated rule: https://github.com/Agilefreaks/Test/pull/42/commits/93c1f7422dca6f49abdf0fbf5ea70e3a92198b85

spec/problem_two_spec.rb
alex-damian-negru: Misunderstood ` let(:count) { $count += 1 }` in https://tinyurl.com/y7aaxlay. Fixed in https://github.com/Agilefreaks/Test/pull/42/commits/bd698d0569424006d3abdafe4524b454c4c5f78c

spec/problem_one_spec.rb
alex-damian-negru: Applied pattern in https://github.com/Agilefreaks/Test/pull/42/commits/72b32fa489ed7a686c5f8be56414c9e6a9209f68.

problem_three.rb
alex-damian-negru: Thanks!

spec/problem_three_spec.rb
alex-damian-negru: Replaced `let` with `subject`: https://github.com/Agilefreaks/Test/pull/42/commits/10b9c0be73f2506285ba04e3212e69b20948885b. Code looks better. 😃

problem_three.rb
alex-damian-negru: Attempted to remove one `return`. Will need to look more into the concept of short-circuiting: https://github.com/Agilefreaks/Test/pull/42/commits/a8f6aaf8a810e864115baaa713370dd75d9a97d4.

problem_three.rb
alex-damian-negru: Is this https://github.com/Agilefreaks/Test/pull/42/commits/a91893176e815122034ec5431e17e1f60d9aa647 version closer to the picture on the left?

problem_one.rb
alex-damian-negru: I tried looking for a more elegant approach and found one at http://ruby-doc.org/core-2.5.1/Enumerable.html#method-i-inject. Used ternary operator instead of `|| 0`: https://github.com/Agilefreaks/Test/pull/42/commits/72ffffbdcdad27b5697dde8b9673c0b635e9a3c6.

problem_one.rb
```diff
@@ -0,0 +1,8 @@
+# Given an integer array of size N, compute the sum of all even numbers in this array.
+
+class ProblemOne
+  def sum_even(array)
+    sum = array.select(&:even?).reduce(:+)
+    sum.nil? ? 0 : sum
```
bunea: The ternary operator still does ` || 0` but in a longer form. [reduce](https://ruby-doc.org/core-2.4.0/Enumerable.html#method-i-reduce) provides a much more elegant way to solve this.

spec/problem_one_spec.rb
```diff
@@ -0,0 +1,33 @@
+require_relative '../problem_one'
+
+describe ProblemOne do
+  it 'computes for a set of even numbers' do
+    even_set = [2, 8, 12, 10, 128, 64, 32]
```
bunea: Could you use `let` in your tests?

Recommended:
https://www.bignerdranch.com/blog/rspec-tip-let-and-nested-describecontext-blocks/
https://relishapp.com/rspec/rspec-core/v/3-8/docs/subject/explicit-subject

bunea (changes requested):

problem_one.rb
levelboy: :+1:

problem_one.rb
```diff
@@ -0,0 +1,8 @@
+# Given an integer array of size N, compute the sum of all even numbers in this array.
+
+class ProblemOne
```
levelboy: Do you need classes for this? What is the need for a class, that you'll then have to instantiate just to call a method on it?

problem_two.rb
```diff
@@ -0,0 +1,7 @@
+# Given an integer array of size N, identify the largest 2 numbers in this array. Do not sort the array.
+
+class ProblemTwo
+  def max_two(array)
+    array.select{ |x| x.is_a? Integer }.max(2)
```
levelboy: Where is the test that made that check required?

spec/problem_one_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    even_sum = subject.sum_even(even_set)
+    expect(even_sum).to eq(256)
+  end
+
+  it 'computes for a set of odd numbers' do
+    odd_set = [1, 5, 11, 3, 1337, 9]
+    odd_sum = subject.sum_even(odd_set)
```
levelboy: Why is this an odd_sum? shouldn't it be the sum of even numbers?

spec/problem_one_spec.rb
```diff
@@ -0,0 +1,33 @@
+require_relative '../problem_one'
+
+describe ProblemOne do
+  it 'computes for a set of even numbers' do
```
levelboy: This is for all the specs. 
It should be clear what method you are testing, what is the context that you are testing for, and what is the actual expectation.
Check RSpec `describe` method, `context` method.

levelboy: This is available for all specs

levelboy (changes requested):
I have more of a general feedback:
1. Why did you choose to use classes and instance methods? What other options do you have?
2. Commit messages are not that descriptive
image
3. I'd say TDD was not correctly understood. For a very good demo, take a look at this: 
Uncle Bob - The Three Laws of TDD - In Kotlin - Prime numbers -> https://www.youtube.com/watch?v=qkblc5WRn-U

Also, a thing to ponder at: What is the difference between Test First Development and Test Driven Development?

problem_three.rb
```diff
@@ -0,0 +1,21 @@
...
+# - Do not use 2 nesting for loops since the time complexity would be high.
+# - You are allowed to destroy the array if you like.
+# - Try to work within the array without using any other temporary data structures.
+# - Try to find a solution by executing just one loop.
+
+class ProblemThree
+  def duplicates?(array)
+    has_duplicate = false
+    array.each_with_index do |_, index|
+      if array[array[index - 1].abs - 1] < 0
```
alexandru-calinoiu: I just love the expressiveness of ruby, did you know you could use `.negative?`

problem_three.rb
```diff
@@ -0,0 +1,21 @@
...
+        has_duplicate = true
+        break
+      else
+        array[array[index - 1].abs - 1] *= -1
```
alexandru-calinoiu: I would avoid duplication by extracting `array[index - 1].abs - 1` into a variable

problem_two.rb
alexandru-calinoiu: Furthermore if you had letters in the array for example `['a', 'b', 'c']` will it work without your `is_a?` check?

alexandru-calinoiu: Using `max` is risky, one criteria was that you should not sort the array, are you sure max does not actually sort the array in the background?

spec/problem_one_spec.rb
alex-damian-negru: It is. The test was meant to handle the case where the result is `0`, as the set contains no even numbers. The test failed in https://github.com/Agilefreaks/Test/pull/42/commits/8adf20673b7f88057108882b0fe11a48576d0650 (RED), which led to handling `0` as a case https://github.com/Agilefreaks/Test/pull/42/commits/ba0cc510bd226148c97dcece41722c8ae97f8545 (GREEN).

`empty_sum` was added afterwards, but it did not undergo the RED-stage, meaning that was a TDD error on my behalf. Will remove `empty_sum` as it is irrelevant and also improve the naming.

problem_one.rb
alex-damian-negru: I actually asked myself the same question! Most of the example tests found were done on methods inside classes, which led me to seeing this as a requirement, although instantiating methods in those three cases seemed somewhat redundant.

Further research proved classes can be avoided in my case using `describe '#method_name'`. 
Will refactor.

spec/problem_one_spec.rb
alex-damian-negru: I'm having a tough time using `let`, `subject` and the three-phase pattern after placing the functions outside classes. Do they all have to be used in a test?

Three-phase pattern: https://www.justinweiss.com/articles/writing-better-tests-with-the-three-phase-pattern/

problem_two.rb
alex-damian-negru: > Furthermore if you had letters in the array for example `['a', 'b', 'c']` will it work without your `is_a?` check?

It would work if we would remove the `select` check, which makes sure only numbers are taken into account, as @drqCode pointed out in https://github.com/Agilefreaks/Test/pull/41, provided the input consists of letters only. `max` returns the letters based on their ASCII values, meaning `max(2) => ['c', 'b']`.

Without the check it would not, however, work on a mixed array such as `['a', 'b', 'c', 100, 101, 99]`.

> Using `max` is risky, one criteria was that you should not sort the array, are you sure max does not actually sort the array in the background?

Based on the [source code ](https://ruby-doc.org/core-2.5.1/Enumerable.html#method-i-max)of the method and further research, no hint was found of `max` sorting the array, making it safe for use & an elegant solution.

spec/problem_one_spec.rb
```diff
@@ -4,34 +4,28 @@
...
       odd_set = [1, 5, 11, 3, 1337, 9]
-      odd_sum = sum_even(odd_set)
-      expect(odd_sum).to eq(0)
-    end
-
-    it 'computes for an empty sum' do
-      empty_set = []
-      empty_sum = sum_even(empty_set)
-      expect(empty_sum).to eq(0)
+      sum = sum_even(odd_set)
```
levelboy, line 27: :+1:

levelboy (approved):

alexandru-calinoiu:
Thank you, closing this as I will followup with an email soon.

PR #43, opened 2018-12-10
Final release

AndreiBoariu:
Final release


AgileFreaks.playground/Contents.swift
```diff
@@ -0,0 +1,136 @@
...
+    func sumIntegersFromArray(arrObjects: [Any]) -> String {
+        var sum = 0
+        
+        if arrObjects.count == 0 {
+            return "Array is empty"
+        }
+        
+        for obj in arrObjects {
+            if let iObj = obj as? Int {
+                sum += iObj
```
alexandru-calinoiu, line 15: The firs problem required to sum the even numbers.

AgileFreaks.playground/Contents.swift
```diff
@@ -0,0 +1,136 @@
...
+    }
+}
+
+class MyTestCase: XCTestCase {
+    
+    func testProblem1() {
+        let p1 = Problem1()
+        p1.sumIntegersFromArray(arrObjects: ["a",2.00,3,4,5,6.3,7,8,9])
+
+        XCTAssertTrue(true)
```
alexandru-calinoiu, line 107: This test is not useful, it will always pass.

alexandru-calinoiu (commented):
There are 2 show stoppers
- the tdd requirement was ignored, and the written tests are useless, denotes fact that you have never written or attempted to write a unit test
- the first problem has the wrong solution.

PR #45, opened 2019-01-31
Test solution and unit tests

rMeleanca:
Here are my solutions to the 3 problems presented, together with the requested unit tests.


Project1/Problem1.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace Problems
+{
+    public class Problem1
+    {
+        public long? SumOfEvenNumbers(int[] numbers)
+        {
+            long? sum = null;
+            if (numbers.Length != 0)
+            {
+                sum = 0;
```
drqCode: Why did you initialized the sum with null? Could the sum to be null?

rMeleanca: I wanted to make a distinction between 2 scenarios:
1. When there are no numbers to be added to the sum, there is no sum, resulting in sum = null
2. When the sum is indeed 0 (eg. adding 2 to -2).

If I leave the sum to be 0 even when there are no numbers to be added to it, it can be interpreted that the two scenarios are the same, when they are not.

rMeleanca: Adding a clarification to scenario 1:
When there are no numbers _in the given array (N=0), there can be_ no sum, resulting in sum = null

drqCode: But this is an assumption... I find more intuitive the sum for an empty set of numbers to be 0, even by a mathematical reason and I couldn't thinks of a case when you want to differentiate between 2 scenarios. Indeed, null is different from 0, but only in 95-99 percent of the problems. In this case, the client code would be required to do null handling...

rMeleanca: That is indeed the better approach. There are compelling mathematical arguments for the "empty sum is zero" convention, and so, I will modify the solution and tests accordingly.

Project1/Problem2.cs
```diff
@@ -0,0 +1,27 @@
...
+                {
+                    max2 = max1;
+                    max1 = number;
+                }
+                else if (max2 == null || number > max2)
+                {
+                    max2 = number;
+                }
+            }
+            return new Tuple<int?, int?>(max1, max2);
```
alexandru-calinoiu, line 24: :+1: for the use of Tuple

Project1/Problem1.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Problems
+{
+    public class Problem1
+    {
+        public long SumOfEvenNumbers(int[] numbers)
```
alexandru-calinoiu, line 7: This is so old school, can you come up with a simpler solution that uses linq?

UnitTestProject1/Problem1Tests.cs
```diff
@@ -0,0 +1,196 @@
...
+        {
+            int[] arr = { 1, 3, 52241, 1201, 11 };
+
+            long resultingSum = problem.SumOfEvenNumbers(arr);
+
+            Assert.AreEqual(resultingSum, 0);
+        }
+
+        [TestMethod]
+        public void SumOfEvenNumbers_AllPositiveNoEvenNumbersDuplicates_SumIsZero()
```
alexandru-calinoiu: This test has no value, there is no point in testing duplicated even numbers.

UnitTestProject1/Problem1Tests.cs
```diff
@@ -0,0 +1,196 @@
...
+        {
+            int[] arr = { 4, 28, 0, 60, 8 };
+
+            long resultingSum = problem.SumOfEvenNumbers(arr);
+
+            Assert.AreEqual(resultingSum, 100);
+        }
+
+        [TestMethod]
+        public void SumOfEvenNumbers_AllPositiveNoOddNumbersDuplicates_SumIs40()
```
alexandru-calinoiu: Again this is redundant with the above test.

alexandru-calinoiu (commented):
Over all this looks nice and clean, I like the structure of the tests.

I make a few comments, the biggest is that many of the tests are redundant.

Project1/Problem1.cs
rMeleanca: Yes. I have made a new change at 974bca6, where a LINQ method syntax is used instead.

UnitTestProject1/Problem1Tests.cs
rMeleanca: I had another look at the tests and removed those that checked for duplicates when there was no need to. Left one that encompasses the entire _duplicates_ scenarios.

`public void SumOfEvenNumbers_PositiveAndNegativeOddAndEvenDuplicates_SumIsMinus80()`

UnitTestProject1/Problem1Tests.cs
rMeleanca: As with the previous case of redundant tests for duplicated numbers, I've made a few changes at 6703401 to account for the observation.

levelboy:
Looks good. Followed up via email. Thank you.

PR #46, opened 2019-02-15
Rezolvari

mhitza:
Pentru problema3 nu am mai vazut pas de refactorizare astfel a fost omis.


spec/problem3_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+  it 'returns true on arrays with duplicates' do
+    expect(problem3([1,2,2])).to eq true
+    expect(problem3([5,4,4,2,1])).to eq true
+  end
+
+  it 'returns false on arrays with no duplicates' do
+    expect(problem3([])).to eq false
+    expect(problem3([1,2,3])).to eq false
+    expect(problem3([5,4,3,2,1])).to eq false
+  end
```
pirvudoru, line 26: What if you add [4,4,1,1] as a test case ?

spec/problem3_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'problem3'
+describe 'problem3' do
+
+  it 'returns true on arrays with duplicates' do
+    expect(problem3([1,2,2])).to eq true
+    expect(problem3([5,4,4,2,1])).to eq true
```
pirvudoru, line 9: In unit tests, it's preferable to use a single expectation per test http://www.betterspecs.org/. Can also make use Rspec's shared examples

lib/problem1.rb
```diff
@@ -0,0 +1,3 @@
+def problem1(values)
+  values.reduce(0) { |sum, value| value % 2 == 0 ? sum + value : sum }
```
alexandru-calinoiu: Ruby's tag line is 'developer pleasure', it has this `?` methods can you search the docs and find one that will be a good fit for this line of code?

lib/problem2.rb
```diff
@@ -0,0 +1,17 @@
+def problem2(values)
+  raise ArgumentError, 'not enough elements for operation' if values.count < 2
+
+  first = nil
+  second = nil
+
+  values.each do |value|
+    if value > first.to_i then
```
alexandru-calinoiu: I guess you are doing `.to_i` because you init values with `nil` the downside is that you will to an unnecessary conversion when a value is populated already. Like haskell or any other modern language ruby has constants for max and min values, can you rewrite this code to make a little more efficient?

lib/problem2.rb
```diff
@@ -0,0 +1,17 @@
...
+      second = first
+      first = value
+    elsif value > second.to_i then
+      second = value
+    end
+  end
+
+  return first, second
```
alexandru-calinoiu: As [rubocop](https://github.com/rubocop-hq/rubocop) will put it: `redundant return statement.`

spec/problem2_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'problem2'
+describe 'problem2' do
+  it 'returns the two largest values from array' do
+    expect(problem2([1,2,3,4,5])).to match_array [4,5]
```
alexandru-calinoiu, line 4: It is generally frowned upon to have multiple assertions in a test, why do you think that is?

spec/problem2_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    expect(problem2([5,4,3,2,1])).to match_array [4,5]
+    expect(problem2([1,5,2,4,3])).to match_array [4,5]
+    expect(problem2([1,4,2,4,3])).to match_array [4,4]
+    expect(problem2([1,1])).to       match_array [1,1]
+  end
+
+  it 'raises ArgumentError when there are\'t two large values' do
+    expect { problem2([1]) }.to raise_error(ArgumentError)
```
alexandru-calinoiu, line 28: Interesting approach, I guess is more of a preference,
I believe that code should flow naturally and exceptions should be used in extreme situations. 
So in this case when the array has less than 2 elements I would prefer to return an empty array. one reason being that if some other code will use the result it will make for a better experience than handling an exception.

Can you give some arguments on why raising an exception is also a good way to go?

alexandru-calinoiu (commented):
Excellent start, made a couple of comments, looking forward to know your thoughts about them.

spec/problem2_spec.rb
mhitza: I wanted to say that from a practical point of view that the failure can be pinpointed from the test "summary" instead of having to review the Failures ("verbose") section of the test run. However can't take credit for that claim any more cause I've opened up the link @pirvudoru shared and it's recommended there with similar justification :)

spec/problem2_spec.rb
mhitza: Given the problem definition, I would consider the function "constrained" to return two values. If it can't it doesn't confine to the expectation of what it's supposed to do, and then it would throw an exception. Handling an exception is generally more cumbersome, I agree, but then the value checks are burdened on the callers of this function and it can become easier (theoretically) to have nil exceptions pop up somewhere else, farther from the source of the code that returned an empty value because it couldn't fulfil it's task.

I do agree though, it's more of a coding preference.

spec/problem3_spec.rb
mhitza: Then I have to rewrite my solution.

mhitza:
I'll work on the changes probably this evening and we can discuss tomorrow on the followup code.

spec/problem2_spec.rb
alexandru-calinoiu: Fair enough, maybe we can continue this discussion some other time.

lib/problem2.rb
mhitza: Was not able to find the constants you where referring to, all I found where user defined constants in stackoverflow examples (or maybe they where taken from ruby source code). And those constants where defined for Fixnum, not Integer, which also contains Bignums.

lib/problem1.rb
mhitza: I've replaced the modulo operator use with the `even?` method

spec/problem3_spec.rb
mhitza: I'll take time to read the shared examples section, because from a quick glance I'm not yet sure how they would be applied in this case

lib/problem2.rb
alexandru-calinoiu: Indeed the MAX and MIN constant have not been yet released, there [is a way](https://gist.github.com/pithyless/9738125) to get those constants now.

Looking at your solution you actually don't need the constants any, do you? You can simply assign the variables initially with the first 2 values of the array, of course you need to compare them first.

One intresting ruby like solution will be something like

```ruby
first, second = values.first(2)
first, second = second, first if second > first
```

What do you think, can you fill in the rest of the code?

spec/problem2_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'problem2'
+describe 'problem2' do
+  it 'returns the two largest values from sorted array' do
+    expect(problem2([1,2,3,4,5])).to match_array [4,5]
+  end
+
+  it 'returns the two largest values from reverse sorted array' do
```
alexandru-calinoiu, line 7: I just now noticed that you don't have a test for negative numbers, it would have been more relevant that this one.

lib/problem3.rb
```diff
@@ -0,0 +1,12 @@
+def problem3(values)
+  incremental_product = 1
+  index_counter = 1
+  array_product = 1
+  while value = values.shift do
+    array_product *= value
```
drqCode, line 6: Please check: https://www.calculatorsoup.com/calculators/discretemathematics/factorials.php for calculating factorials. Given you have N numbers from 1 to N. What is the biggest N you can have for your algorithm to run correctly?

spec/problem3_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'problem3'
+describe 'problem3' do
```
drqCode, line 2: please add the test for this input `[2, 2, 2, 3]`. Does it pass?

lib/problem3.rb
```diff
@@ -0,0 +1,12 @@
...
+    incremental_product *= index_counter
```
drqCode, line 7: You can get rid of this statements and related ones to it. As you receive an array, you have the size of it, so you may get the factorial of it. See: https://stackoverflow.com/questions/2434503/ruby-factorial-function

mhitza:
Well I'm out of ideas for problem3 given the constraints of the solution. I think we can consider it failed.

alexandru-calinoiu:
Closing this due to lack of followup. It was fun while it lasted :)

PR #47, opened 2019-03-15
George Bancila test pull request

ghost:


problems/prob1.rb
```diff
@@ -0,0 +1,10 @@
+#Given an integer array of size N, compute
+#the sum of all even numbers in this array
+
+def sum_of_even_numbers(numbers_array)
+  if numbers_array.nil? || numbers_array.empty? || numbers_array.any? { |e| e == nil  }
```
drqCode: Mathematically, sum of the elements of an empty set is 0, not nil. I think this is more naturally. Also, given the description of the problem, you won't need nil checks. The client code should receive an error if it doesn't use the function correctly

problems/prob1.rb
```diff
@@ -0,0 +1,10 @@
...
+    return nil
```
drqCode: You can omit `return`. Check [this](https://medium.com/rubycademy/the-return-keyword-in-ruby-df0a7f578fcb)

problems/prob2.rb
```diff
@@ -0,0 +1,15 @@
...
+#Do not sort the array.
+
+def largest_two_numbers(numbers)
+  if numbers.nil? || numbers.empty? || numbers.any? { |e| e == nil  }
+    return nil
+  elsif numbers.size == 1
+      return numbers
+  else
+    largest = numbers.max()
+    numbers.delete_at(numbers.index(largest))
```
drqCode: What issue is possible if the `numbers` array has a higher scope? `delete_at` mutates the array...

problems/prob3.rb
```diff
@@ -0,0 +1,24 @@
+#Given an array of numbers having size N in which every number is between 1 and N,
+#determine if there are any duplicates in it. You are allowed to destroy the array if you like.
+#Do not sort the array or use bit vectors. Try to work within the array without using any other temporary
+#data structures. Do not use 2 nesting for loops since the time complexity would be high.
+#Try to find a solution by executing just one loop.
+#Hint: manipulate the elements of the same array as you loop through.
+
+def check_duplicates(numbers)
```
drqCode: If you payed attention to `#nil?` and `#empty?` methods, they have `?` at end of the name, meaning the methods return `boolean`. This is a ruby convention for such methods. Could you use that too?

drqCode (commented):
Good solutions to the problems, but some improvements regarding ruby are needed.

problems/prob2.rb
ghost: I did it like that because in the problem description it didn't said that I was not allowed to destroy the array.

drqCode: Yes, but for getting 2 largest numbers, it is not justified enough, intuitively...

problems/prob1.rb
```diff
@@ -0,0 +1,10 @@
...
+  else
+    return numbers_array.select{|x| x % 2 == 0}.sum
```
alexandru-calinoiu: This is almost perfect, but ruby makes this __even__ nicer, check the the integere [documentation](https://ruby-doc.org/core-2.6.2/Integer.html) and see if you notice something usefull there.

problems/prob2.rb
```diff
@@ -0,0 +1,15 @@
+#Given an integer array of size N, identify the largest 2 numbers in this array.
+#Do not sort the array.
+
+def largest_two_numbers(numbers)
+  if numbers.nil? || numbers.empty? || numbers.any? { |e| e == nil  }
```
alexandru-calinoiu: This is not needed, but if it was needed is considered a code smell to repeat the same snippet of code.

Now that you have this around can you [dry'it](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

Goes to prove that if you write more code than needed it will generate even more workload :)

tests/prob1-tests.rb
```diff
@@ -0,0 +1,34 @@
+require 'test/unit'
+require_relative "../problems/prob1"
+
+class UnitTests < Test::Unit::TestCase
```
alexandru-calinoiu, line 4: Tests look very nice, they have proper structure and they mostly test the right thing, a noticeable exception is that nil check.

alexandru-calinoiu (changes requested):
Thank you for you submission, is a promising start, could you please review comments and see if you can act on any?

problems/prob1.rb
ghost:
```suggestion
    return numbers_array.select{|x| x.even?}.sum
```

problems/prob1.rb
ghost: I refactored all the problems with this idea. Didn't knew Ruby could do that

problems/prob2.rb
ghost: Constantin suggested that I should not check for nil given the problem description and I removed some parts in the latest commit.

alexandru-calinoiu: Indeed that is better,

I was just making a point that more code, even if it's necessarily good code, will open the way for issues.

problems/prob1.rb
```diff
@@ -0,0 +1,10 @@
+#Given an integer array of size N, compute
+#the sum of all even numbers in this array
+
+def sum_of_even_numbers(numbers_array)
+  if numbers_array.empty?
```
alexandru-calinoiu: This is nitpicking, but after you will write some more ruby I bet you will agree.

Open your favourite terminal and type `irb`, you should get a nice prompt where you can type ruby code and press enter to run it, type there `[].sum` and notice what happens, come back to this problem and take a second look at it ;)

problems/prob2.rb
```diff
@@ -0,0 +1,28 @@
+#Given an integer array of size N, identify the largest 2 numbers in this array.
+#Do not sort the array.
+
+def largest_two_numbers(numbers)
+  if numbers.empty?
+    0
```
alexandru-calinoiu: The return is not consistent, if you sometime return a single integer and sometime an array will confuse the caller a lot.

problems/prob1.rb
ghost: I refactored the problem in the latest commit.

alexandru-calinoiu:
Thank you, will close this and will followup via email.

problems/prob1.rb
drqCode: You can learn a lot of things regarding ruby by using [rubocop](https://github.com/rubocop-hq/rubocop).

PR #48, opened 2019-04-21
Done with all three problems and unit tests for them

CristinaBirsan:


duplicate.js
```diff
@@ -0,0 +1,34 @@
...
+     return false;
+  }
+  else {
+      return true;
+  }
+}
+
+var unitTest={};
+
+unitTest.testDupl = function(method){
```
alexandru-calinoiu: This is not the way unit tests are usually written in javascript, please search for a common way and adjust your implementation.

alexandru-calinoiu (changes requested):
Please review my comments, we need proper unit testing before we start looking at the implementations more closely.

CristinaBirsan:
Trebuie sa folosesc un framework pentru unit tests?

suma.js
```diff
@@ -0,0 +1,26 @@
+function sum(arr)
+{
+  sum = 0;
+  for(var i =0 ; i<arr.length;i++){
+    sum+=arr[i];
```
alexandru-calinoiu: The problem stated `Given an integer array of size N, compute the sum of all *even* numbers in this array`

Is even numbers not all numbers.

max2.js
```diff
@@ -0,0 +1,42 @@
+function secondMax(arr){
+  var max1 = Math.max.apply(null,arr);
+  var max2 =Math.max.apply(null,arr);
```
alexandru-calinoiu: This computation of max is a waste of time, as it will be overwritten later on.

You probably want to remove `max1` from the array and the compute max2.

max2.js
```diff
@@ -0,0 +1,42 @@
+function secondMax(arr){
```
alexandru-calinoiu: Even when corrected this solution is inefficient, we can suppose `Math.max` will traverse the entire array, in your current implementation this happens 3 times.

Can you come up with a solution that only traverses the array 1 time.

duplicate.js
```diff
@@ -0,0 +1,34 @@
+function duplicate(arr){
+  var a =new Set(arr);
```
alexandru-calinoiu: Again a miss read of the problem statement `Try to work within the array without using any other temporary data structures.`

The set here ([poorly](https://blog.elpassion.com/naming-101-quick-guide-on-how-to-name-things-a9bcbfd02f02) named a) is a temporary data structure.

alexandru-calinoiu (changes requested):
Using a custom unit testing framework is a interesting solution and very unusual, it might be worth exploring this further given that it can be an overhead to add a testing framework.

The code is poorly formatted and and not following any coding conventions.
The solutions for 2 of the problems don't respect specified constrains.

suma.js
CristinaBirsan: I will restore the code and send it as soon as possible.

test/duplicateTest.js
```diff
@@ -0,0 +1,10 @@
+var assert= require('chai').assert;
+
+var duplicate = require('../duplicate');
+
+describe('Duplicate', function(){
+    it("should return True(Duplicate)", function(){
+        var result = duplicate([1,2,3,2,4]);
+        assert.equal(result,true);
+    });
+});
```
alexandru-calinoiu, line 10: We should have unit test for other use cases also.

duplicate.js
```diff
@@ -0,0 +1,8 @@
+module.exports = function(arr){
+    for(var i = 0 ; i<arr.length;i++){
+        if(arr.indexOf(arr[i])!=i){
```
alexandru-calinoiu, line 3: This is a improved solution indeed.

[indexOf](https://www.w3schools.com/jsref/jsref_indexof_array.asp) will search the array via an loop, this means that you will have an inner loop.

The problem stated that you should not use a inner loop.

test/twoMaxTest.js
```diff
@@ -0,0 +1,11 @@
+var assert= require('chai').assert;
+
+var towMax = require('../twoMax');
+
+describe('1 & 2 Max Element', function(){
+    it("should return 200 and 80", function(){
+        var result = towMax([10,200,80,15]);
+        var testResult= '200,80';
+        assert.equal(result,testResult);
+    });
```
alexandru-calinoiu, line 10: If you write a test with an array like `[-5, -4]` it will fail as you don't account for negative numbers

twoMax.js
```diff
@@ -0,0 +1,17 @@
...
+        maxFirst=arr[i];
+      }
+      else if(arr[i]>maxSecond){
+        maxSecond=arr[i];
+      }
+    } 
+    var v= new Array();
+    v[0]=maxFirst;
+    v[1]=maxSecond;
+    return v.toString();
```
alexandru-calinoiu, line 16: Why return a string vs an array?

alexandru-calinoiu (commented):
This is an improvement over the previous implementation, left some comments there is nothing to act on now.

Will close the pr and get back to you via email.

PR #50, opened 2020-11-22
My solution to problems 1, 2 and 3

scay0x:
Some mentions for problem number 3:
- I resolved the problem in multiple ways, because I was not sure about each proposed solution.
- However I still proposed a solution with nested fors, although the problem requirement clearly specified to not.

### DIFF
--- a/Test1/src/Problem1.kt
+++ b/Test1/src/Problem1.kt
@@ -7,5 +7,5 @@ fun main() {
             sum += arr[i]
         }
     }
-    println(sum)
+    println("suma = " + sum)
 }
--- a/Test1/src/Problem1.kt
+++ b/Test1/src/Problem1.kt
@@ -1,0 +1,11 @@
+fun main() {
+    val N = 10
+    var sum = 0
+    val arr = IntArray(N) { i -> i }
+    for(i in arr) {
+        if (i % 2 == 0) {
+            sum += arr[i]
+        }
+    }
+    println(sum)
+}
--- a/Test1/src/Problem2.kt
+++ b/Test1/src/Problem2.kt
@@ -1,0 +1,22 @@
+fun main() {
+    var max1 = 0
+    var max2 = 0
+
+    val arr = intArrayOf(17,11,10,19,15,23,1,9,100)
+    for(valoare_i in arr) {
+        if(valoare_i > max1) {
+            max1 = valoare_i
+        }else {
+            max1 = max1
+        }
+    }
+    for(valoare_j in arr) {
+        if (valoare_j > max2 && valoare_j != max1) {
+            max2 = valoare_j
+        } else {
+            max2 = max2
+        }
+    }
+    println("max1 = " + max1)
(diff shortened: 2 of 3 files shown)


Test1/src/Problem1.kt
```diff
@@ -0,0 +1,11 @@
+fun main() {
```
alexandru-calinoiu, line 1: The implementation is straight forward

The idea was that you get an array of n elements as an input, you don't have to generate the input.

Test1/src/Problem1.kt
```diff
@@ -0,0 +1,11 @@
...
+    val N = 10
+    var sum = 0
+    val arr = IntArray(N) { i -> i }
+    for(i in arr) {
+        if (i % 2 == 0) {
+            sum += arr[i]
+        }
+    }
```
alexandru-calinoiu, line 9: There a couple of [usefull functions for arrays in kotlin](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-array/) scout the docs maybe you can find a way to make this more idiomatic.

Test1/src/Problem2.kt
```diff
@@ -0,0 +1,22 @@
...
+            max1 = max1
+        }
+    }
+    for(valoare_j in arr) {
+        if (valoare_j > max2 && valoare_j != max1) {
+            max2 = valoare_j
+        } else {
+            max2 = max2
+        }
+    }
```
alexandru-calinoiu, line 19: Can you think of a way to do this in one loop?

Also what about some edge cases? empty array, one item array ....

Test1/src/Problem3.kt
```diff
@@ -0,0 +1,56 @@
...
+fun main() {
+    var a = intArrayOf(1204, 66, 7423, 1, 4, 850, 4, 55, 10003, 55, 78364, 93, 403, 1789, 1, 3850, 778, 850)
+
+    var distincte = ""
+    var duplicate = ""
+
+    for(i in a.indices) {
+        if(i == 0) {
+            distincte += (a[i].toString() + " ")
+        } else if(distincte.contains(" " + a[i])) {
```
alexandru-calinoiu, line 12: `contains` most likely uses a loop in it's implementation, so you are still using a loop inside a loop, there is a clear way to solve this problem.

alexandru-calinoiu (changes requested):
Took a look at the problems, apprecite the fact that you've done them in kotlin, made a couple of commets take a look.

The main problem is that you did not do any unit tests to prove the functionality of any of the solutions.

deniscoman:
Beside what @alexandru-calinoiu said, you have some naming issues that makes your code harder to read.
1. It's better to write all your code in English so anyone can understand what are you trying to do.
2. [Here](https://dzone.com/articles/naming-conventions-from-uncle-bobs-clean-code-phil) you can find some naming conventions that will help you, and anyone else who read your code, to understand what your code does. This will be helpful in the feature no matter what programming language you will use.

